PR #129, opened 2019-01-14
Adding supportedManagedDependencies.json with a list of managed dependencies supported by Azure Functions.

Francisco-Gamino:
Adding supportedManagedDependencies.json file with a list of managed dependencies supported by Azure Functions.

* Managed dependencies are PowerShell modules managed by Azure Functions. At the present time, only the 'Az' module version '1.0.1' is supported.

* Once a managed dependency module is supported in Azure Functions, the module version in this file will be updated.

* This file will be used by the Functions CLI to setup the user's Client machine with the latest supported list of managed dependencies.

### Client UX from CLI + PowerShell Core

When a new Function App is initialized locally (via func init), the managedDependencies setting will be added to the host.json on the local machine for PowerShell. The initialization of the Function App will download the latest Azure (Az) modules and place them in the ManagedDependencies folder in the Function App. When a function is uploaded, the values in host.json is read and this is applied to the Function App. The local modules  in the ManagedDependencies folder are not packaged up and uploaded to the service.

```powershell
<#
Prerequisites: 
    - PowerShell Core
    - Functions CLI
#>

# Start PowerShell Core.
# Make sure the Functions CLI is part of the PATH environment. 

# Create a local directory where the function app will be created
New-Item -Path c:\MyFunctionApplication -ItemType Directory -Force

# Create a function app via func.exe. When running this command,
# the user can opt in to Managed Dependencies by using the --managed-dependencies switch.
func init MyFunctionApplication --worker-runtime powershell --managed-dependencies

<#
# After this, the Az modules are installed under the Function App
c:\MyFunction\Modules
c:\MyFunction\ManagedDependencies\Az
c:\MyFunction\ManagedDependencies\Az\1.0.0
c:\MyFunction\ManagedDependencies\Az.Aks\1.0.0
c:\MyFunction\ManagedDependencies\Az.AnalysisServices\1.0.0
c:\MyFunction\ManagedDependencies\Az.ApiManagement\1.0.0
c:\MyFunction\ManagedDependencies\Az.ApplicationInsights\1.0.0
c:\MyFunction\ManagedDependencies\Az.Automation\1.0.0
c:\MyFunction\ManagedDependencies\Az.Batch\1.0.0
c:\MyFunction\ManagedDependencies\Az.Billing\1.0.0
c:\MyFunction\ManagedDependencies\Az.Cdn\1.0.0
# More Az modules...
#>

# Once the installation of managed dependencies (modules 

# Entry in Host.json for managed dependencies
Host.json file:
    "managedDependencies": [
        {
            "Name": "Az",
            "MajorVersion": "1"           
        }
    ]

```


daxian-dbw:
@Francisco-Gamino I heard from @TylerLeonhardt that this JSON file is used in core-tools. Maybe this JSON file should be a static resource at https://github.com/Azure/azure-functions-core-tools/tree/45f08c695a141c970fabcc6a3ab32628ff5845f8/src/Azure.Functions.Cli/StaticResources. @TylerLeonhardt just submitted a PR to core-tools to add `profile.ps1` as a static resource and create the file on the fly when running `func init --worker-runtime powershell`. I think we should do the same for this JSON file. Thoughts?

Francisco-Gamino:
Thank you @daxian-dbw  for the review. This json file is not going to be a static resource. It will be updated when a new version of the Az modules is available in the service.

The pipeline will look like this:
1) A new version of the Az modules is available in the PSGallery
2) Make that version available in the service (Azure Functions)
3) Once available in the service, update this json file withe the latest supported version

Currently, the Azure PowerShell team releases a new minor version of the Az modules approximately every two weeks.

We decided to put this file in this repo because the Az module is a PowerShell resource. If someday managed dependencies are enable for Python, then an equivalent file will probably be added to that worker repo. 

Please let me know if you have any other questions. Thanks!

daxian-dbw:
@Francisco-Gamino This file will probably be moved to another more suitable repo in future (say, the dependency-rapid-deploy repo have such `json` files for all supported languages). In order to allow us to remove it from this repo once it's moved to a more permanent location, the `func cli` needs to use a forward link to pull this file down on the users' local machine instead of using the static URL to `github.com\...\managedDependencies.json`. Do you agree with that?

PR #153, opened 2019-02-28
Enable installing function app module dependencies.

Francisco-Gamino:
### Added support to install function app dependencies

* This is draft PR, so we can start reviewing the product code. I will work on adding tests next.
Changes summary:
* Updated FuntionRpc.proto to latest version which includes managed dependencies
Updated azure-functions-powershell-worker.sln given that the test project was not being loading. Now this is fixed.
* Added PackageManagement and PowerShellGet modules to the PSWorker. These will be used to download function app dependencies
* Updated Grpc.Core from version 1.14.1 to 1.18.0
* Updated build.ps1 to remove the Modules/PackageManagement/fullclr which contains the same contents as the coreCLR folder.
* Added support to install function app dependencies via Requirements.psd1. 
Requirements.psd1 file content:
@{
    'Az' = '1.*'
}

### Installation logic
During function load request, we check if functionLoadRequest.ManagedDependencyEnabled is enable. If it is, we resolve the function app root and search for Requirements.psd1. If the psd1 file is found, we parse using the PowerShell API. From there, we build a list of DependenyInfo objects which contain all the information about the dependent module name, version, destinationPath, and whether it needs to be installed. Before we create the DependenyInfo object, we do the validate that the module name is a supported managed dependency. Currently, only Az is supported, but this list can be expanded as needed. We also validate that the version is in the correct format (‘majorVersion.*’) and check that the given major version is less or equal to the one available major version in the PSGallery.

After initializing PowerShell, the module are installed to the ManagedDependencies folder under the function app for the local scenario and under the HOME\data directory when running on Azure. Right before we install the module, we do a one last check to see if the module is already installed. We search for the moduleDestinationPath\ModuleName\LatestVersion. If it exist, the module is already installed. Otherwise, we install it using the Save-Module cmdlet. Please note that Save-Module is able to download versions of Az module side-by-size. However, the dependent modules get overwritten with the latest version. As a result, the previous Az version will not longer work. If this is the case, we empty the Az directory before running Save-Module.

If an error is encountered during the function app dependency installation, we wrap the exception in a DependencyInstallationException. Then, in RequestProcessor.cs we unwrapped the exception and set the response response.FunctionLoadResponse.IsDependencyDownloaded to false. This means that we failed to download the function app dependencies. Lastly, response.FunctionLoadResponse.IsDependencyDownloaded is set to true if everything installed with no errors.

Thanks,
Francisco


Francisco-Gamino:
cc: @pragnagopa

Francisco-Gamino:
cc: @Hazhzeng

src/DependencyManagement/DependencyManager.cs
```diff
@@ -0,0 +1,167 @@
...
+            var functionAppRootPath = Path.GetFullPath(Path.Join(request.Metadata.Directory, ".."));
+
+            // Location where the dependent modules will be installed.
+            var modulesFolderPath = Path.Join(functionAppRootPath, "Modules");
+
+            // Path to Requirements.psd1 file.
+            var requirementsFilePath = Path.Join(functionAppRootPath, RequirementsPsd1FileName);
+
+            // Parse and process Requirements.psd1.
+            Hashtable entries = ParsePowerShellDataFile(requirementsFilePath);
```
SatishRanjan: entries need to be checked for nullability before using it as ParsePowerShellDataFile can return null for the Hashtable

src/DependencyManagement/DependencyManagementUtils.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.IO;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
+{
+    internal class DependencyManagementUtils
+    {
+        // File name which holds module information after it is downloaded/installed from the PSGallery.
+        private const string DependencyInfoJsonFileName = "DependencyInfo.json";
```
pragnagopa: What is the purpose of DependencyInfo.json? when you are already building denpendencyinfo objects

src/PowerShell/PowerShellManager.cs
```diff
@@ -72,6 +74,9 @@ internal PowerShellManager(ILogger logger)
             _pwsh.Streams.Verbose.DataAdding += streamHandler.VerboseDataAdding;
             _pwsh.Streams.Warning.DataAdding += streamHandler.WarningDataAdding;
 
+            // Install function app dependent modules
+            initAction?.Invoke(_pwsh, logger);
```
pragnagopa: change this to async

src/DependencyManagement/DependencyManagementUtils.cs
Francisco-Gamino: @pragnagopa : When we install a module (via Save-Module), we create a DependecyInfo.json entry with the name and version of the module that was installed. This way, next time the function app is initialized, we can use this file to try to figure out if we have already installed the module at the requested path. If the module is already there, we check the version to ensure we have the latest one. For reference, please see RunInstallManagedDependencyModule

src/Messaging/Dependency.cs
```diff
@@ -0,0 +1,183 @@
+// <auto-generated>
+//     Generated by the protocol buffer compiler.  DO NOT EDIT!
```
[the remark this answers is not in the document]
Francisco-Gamino: Good catch. Removing generated files.

src/Microsoft.Azure.Functions.PowerShellWorker.csproj
```diff
@@ -35,4 +35,8 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Messaging\" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
Francisco-Gamino: Fixed.

src/PowerShell/PowerShellManager.cs
Francisco-Gamino: This is an synchronous call, so the calling thread has to wait. Will keep it as is.

src/DependencyManagement/DependencyManagementUtils.cs
pragnagopa: Can you expand on what you mean by 

> next time the function app is initialized

We can sync up in person if needed.

protobuf/src/proto/ManagedDependency/Dependency.proto
```diff
@@ -0,0 +1,17 @@
...
+option java_package = "com.microsoft.azure.functions.rpc.messages";
+option java_outer_classname = "DependencyProto";
+option csharp_namespace = "Microsoft.Azure.WebJobs.Script.Grpc.Messages";
+option go_package ="github.com/Azure/azure-functions-go-worker/internal/rpc";
+
+package AzureFunctionsRpcMessages;
+
+//Used to send the managed dependency module for the language worker to download the client module
+message Dependency {
+  //Flag indicating if managed dependency is enabled or not
```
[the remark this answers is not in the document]
Francisco-Gamino: @pragnagopa : Done.

src/DependencyManagement/DependencyManager.cs
Francisco-Gamino: Fixed. Thanks @SatishRanjan

src/DependencyManagement/DependencyManagementUtils.cs
Francisco-Gamino: @pragnagopa : Let's sync in person tomorrow. It would be easier to explain the logic in the whiteboard.

src/RequestProcessor.cs
```diff
@@ -187,8 +240,8 @@ private void ProcessInvocationRequestImpl(StreamingMessage request, AzFunctionIn
         private StreamingMessage NewStreamingMessageTemplate(string requestId, StreamingMessage.ContentOneofCase msgType, out StatusResult status)
         {
             // Assume success. The state of the status object can be changed in the caller.
-            status = new StatusResult() { Status = StatusResult.Types.Status.Success };
```
Francisco-Gamino: Note: Fix file formatting.

src/DependencyManagement/DependencyManagementUtils.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            if (string.IsNullOrWhiteSpace(requestContent))
+            {
+                return null;
+            }
+
+            // Load up the XML response.
+            XmlDocument doc = new XmlDocument();
+            XmlReaderSettings settings = new XmlReaderSettings();
```
SatishRanjan: Please remove reader setting as this is not used for XmlReader

src/DependencyManagement/DependencyManagementUtils.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            string requestContent = null;
+
+            // Get the response response.
+            using (HttpWebResponse response = request?.GetResponse() as HttpWebResponse)
+            {
+                var stream = response?.GetResponseStream();
+                if (stream != null)
+                {
+                    StreamReader reader = new StreamReader(stream);
```
SatishRanjan: This will require retry to handle the scenario of web request failure

src/DependencyManagement/DependencyManager.cs
```diff
@@ -0,0 +1,223 @@
...
+        // Requirements.psd1 file name.
+        private const string RequirementsPsd1FileName = "requirements.psd1";
+
+        // The list of managed dependencies supported in Azure Functions.
+        internal static readonly List<string> SupportedManagedDependencies = new List<string>() {"Az"};
+
+        // Environment variables to help figure out if we are running in Azure.
+        private const string AzureWebsiteInstanceId = "WEBSITE_INSTANCE_ID";
+        private const string HomeDriveName = "HOME";
+        private const string dataFolderName = "data";
```
SatishRanjan: per convention of naming constant variable it should start with capital case, so please rename this to "DataFolderName"

src/DependencyManagement/DependencyManager.cs
```diff
@@ -0,0 +1,223 @@
...
+        internal void SetFunctionAppDependencies(FunctionLoadRequest request)
+        {
+            // Resolve the FunctionApp root path.
+            var functionAppRootPath = Path.GetFullPath(Path.Join(request.Metadata.Directory, ".."));
+
+            // Resolve the managed dependencies folder path.
+            DependenciesPath = GetManagedDependenciesPath(functionAppRootPath);
+
+            // Parse and process Requirements.psd1.
+            Hashtable entries = ParsePowerShellDataFile(functionAppRootPath, RequirementsPsd1FileName);
```
SatishRanjan: Please remove RequirementsPsd1FileName, as the version is part of path the installed module

src/DependencyManagement/DependencyManager.cs
Francisco-Gamino: Fixed.

src/DependencyManagement/DependencyManagementUtils.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <summary>
+        /// Deletes the contents at the given directory.
+        /// </summary>
+        internal static void EmptyDirectory(string path)
+        {
+            var directoryInfo = new DirectoryInfo(path);
+
+            if (directoryInfo.Exists)
+            {
+                foreach (var file in directoryInfo.GetFiles())
```
Hazhzeng: Is it possible to remove the directory recursively if exist, and create again. In this case, we don't need to do a forloop.
```
if (directoryInfo.Exists) {
    directoryInfo.Delete(true);
}

directoryInfo.Create();
```

Francisco-Gamino: @Hazhzeng : Good point. However, there is a known issue sporadically when the directory is not empty. This is documented here: https://stackoverflow.com/questions/329355/cannot-delete-directory-with-directory-deletepath-true

src/RequestProcessor.cs
```diff
@@ -281,6 +337,151 @@ private void BindOutputFromResult(InvocationResponse response, AzFunctionInfo fu
...
+                var errorMsg = string.Format(PowerShellWorkerStrings.FailToInstallFuncAppDependencies);
+                var dependencyInstallationException = new DependencyInstallationException(errorMsg, e);
+                throw dependencyInstallationException;
+            }
+        }
+
+        /// <summary>
+        /// Installs managed dependencies modules specified in Requirements.psd1 for the function app.
+        /// </summary>
+        private void RunInstallManagedDependencyModule(System.Management.Automation.PowerShell pwsh, ILogger logger)
```
SatishRanjan: This needs to be made testable by not having this as private method, as there're lots of logic inside this function

src/RequestProcessor.cs
```diff
@@ -281,6 +337,151 @@ private void BindOutputFromResult(InvocationResponse response, AzFunctionInfo fu
...
+                    .AddParameter("RequiredVersion", version)
+                    .AddParameter("path", path)
+                    .AddParameter("Force", true)
+                    .AddParameter("ErrorAction", "Stop")
+                    .InvokeAndClearCommands();
+            }
+            catch (Exception e)
+            {
+                exception = e;
+                throw;
```
daxian-dbw: I think this exception should be made an inner exception of the dependency exception that you will throw, and the message from this exception should be incorporated into the message of the outer exception.
So you don't need to catch it here.

src/RequestProcessor.cs
```diff
@@ -281,6 +337,151 @@ private void BindOutputFromResult(InvocationResponse response, AzFunctionInfo fu
...
+            }
+            finally
+            {
+                if (pwsh.HadErrors)
+                {
+                    string errorMsg = string.Format(PowerShellWorkerStrings.FailToInstallModule, moduleName, version);
+                    logger.Log(LogLevel.Error, errorMsg, exception, isUserLog: true);
```
daxian-dbw: Better to move this to the catch block.

src/RequestProcessor.cs
```diff
@@ -281,6 +337,151 @@ private void BindOutputFromResult(InvocationResponse response, AzFunctionInfo fu
...
+                }
+                else
+                {
+                    var message = string.Format(PowerShellWorkerStrings.ModuleHasBeenInstalled, moduleName, version);
+                    logger.Log(LogLevel.Trace, message, isUserLog: true);
```
daxian-dbw: better to move this to the end of try block, so it's done when `pwsh.invoke` is successful.

src/RequestProcessor.cs
```diff
@@ -281,6 +337,151 @@ private void BindOutputFromResult(InvocationResponse response, AzFunctionInfo fu
...
+            string moduleName,
+            string version,
+            string path)
+        {
+            Exception exception = null;
+
+            // If the destination path does not exist, create it.
+            if (!Directory.Exists(path))
+            {
+                Directory.CreateDirectory(path);
```
SatishRanjan: These can be abstracted using Package System.IO.Abstractions

https://github.com/System-IO-Abstractions/System.IO.Abstractions

src/DependencyManagement/DependencyManagementUtils.cs
Hazhzeng: @Francisco-Gamino Thanks for clarification 👍

src/DependencyManagement/DependencyManager.cs
```diff
@@ -0,0 +1,223 @@
...
+            return hashtable;
+        }
+
+        /// <summary>
+        /// Validates the given version format. Currently, we only support 'Number.*'.
+        /// </summary>
+        internal bool IsValidVersionFormat(string version)
+        {
+            var pattern = @"^(\d){1,2}(\.)(\*)";
+            var regex = new Regex(pattern);
```
daxian-dbw: Use `Regex.IsMatch` instead.

src/PowerShell/PowerShellManager.cs
```diff
@@ -23,6 +23,8 @@ internal class PowerShellManager
         private readonly ILogger _logger;
         private readonly PowerShell _pwsh;
 
+        private const string ProfileFileName = "profile.ps1";
```
daxian-dbw: No need to have this const string. It's not used anywhere else. You can directly use it below.

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -1,166 +1,208 @@
-<?xml version="1.0" encoding="utf-8"?>
-<root>
```
daxian-dbw: Can you please reformatting this file to make it less different from the original? Spaces characters were added to many places and cause it to look so differently. It's hard to review changes in this file.

protobuf/src/proto/FunctionRpc.proto
```diff
@@ -212,6 +215,9 @@ message FunctionLoadResponse {
   // Result of load operation
   StatusResult result = 2;
   // TODO: return type expected?
+
+  // Result of load operation
+  bool is_dependency_downloaded = 3;
```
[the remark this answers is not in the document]
Francisco-Gamino: @SatishRanjan : Could you please explain how is this being used in the Host (runtime)?

src/RequestProcessor.cs
Francisco-Gamino: Yes, it is a TODO item to fix the formatting.

src/DependencyManagement/DependencyManager.cs
Francisco-Gamino: Fixed. The code looks like this now:

var pattern = @"^(\d){1,2}(\.)(\*)";
return Regex.IsMatch(version, pattern);

src/PowerShell/PowerShellManager.cs
Francisco-Gamino: Fixed.

src/DependencyManagement/DependencyManagementUtils.cs
Francisco-Gamino: @Hazhzeng : Thank you for taking a look!

src/DependencyManagement/DependencyManager.cs
Francisco-Gamino: Fixed. The code looks like this now:
var pattern = @"^(\d){1,2}(.)(*)";
return Regex.IsMatch(version, pattern);

protobuf/src/proto/FunctionRpc.proto
```diff
@@ -202,6 +202,9 @@ message FunctionLoadRequest {
 
   // Metadata for the request
   RpcFunctionMetadata metadata = 2;
+
+  // A flag indicating if managed dependency is enabled or not
+  bool managed_dependency_enabled = 3;
```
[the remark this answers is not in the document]
Francisco-Gamino: @daxian-dbw : This PR has already been approved and merged to the dev branch of the protobuf repo. @SatishRanjan completed this work.

build.ps1
```diff
@@ -77,8 +77,24 @@ if(!$NoBuild.IsPresent) {
...
         -OutFile "$PSScriptRoot/src/Modules/Microsoft.PowerShell.Management/Microsoft.PowerShell.Management.psd1" 
 
+    # Compile project
+    Write-Log "Compiling project..."
     dotnet publish -c $Configuration $PSScriptRoot
+    if ($LASTEXITCODE -ne 0) { throw "Build compilation failed." }
+
+    # Remove the fullCLR folder under PackageManagement (since it has the same content at the coreclr folder)
+    $packageManagementFullCLRFolderPath = (Get-Item "$PSScriptRoot/src/bin/$Configuration/*/Modules/PackageManagement/1.1.7.0/fullclr" -ErrorAction SilentlyContinue).FullName
+    if (![string]::IsNullOrWhiteSpace($packageManagementFullCLRFolderPath) -and (test-path $packageManagementFullCLRFolderPath))
```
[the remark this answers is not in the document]
Francisco-Gamino: I agree. Fixed.

build.ps1
```diff
@@ -77,8 +77,24 @@ if(!$NoBuild.IsPresent) {
...
+    {
+        Write-Log "Deleting PackageManagement/1.1.7.0/fullclr folder..."
+        Remove-Item $packageManagementFullCLRFolderPath -Force -Recurse -ErrorAction Stop
+        Write-Log "    Completed."
+    }
```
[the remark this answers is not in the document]
Francisco-Gamino: Fixed.

src/RequestProcessor.cs
Francisco-Gamino: Really good point. Fixed.

protobuf/src/proto/FunctionRpc.proto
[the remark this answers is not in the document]
SatishRanjan: Host has to log if the function execution failure is caused by the dependency download request.

protobuf/src/proto/FunctionRpc.proto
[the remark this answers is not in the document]
daxian-dbw: Yes, I know it's merged to the protobuf repo. But updating the copies in the worker repo requires to be done in a separate PR, as shown in that doc.

protobuf/src/proto/FunctionRpc.proto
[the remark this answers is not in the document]
daxian-dbw: Got it. Thanks

src/DependencyManagement/DependencyManagementUtils.cs
Francisco-Gamino: @SatishRanjan : I am updating this code with your latest iteration. Let me know if there is anything you would like me to update here before pushing the code upstream.

src/FunctionLoader.cs
```diff
@@ -51,20 +49,26 @@ internal void LoadFunction(FunctionLoadRequest request)
...
             var workerLevelModulesPath = Path.Join(AppDomain.CurrentDomain.BaseDirectory, "Modules");
-            FunctionModulePath = $"{appLevelModulesPath}{Path.PathSeparator}{workerLevelModulesPath}";
-            if (!string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(AzureWebsiteInstanceId))
-                && Platform.IsWindows)
+
+            // Add the managed dependencies folder path
+            if (!string.IsNullOrWhiteSpace(managedDependenciesPath))
             {
-                FunctionModulePath = $"{FunctionModulePath}{Path.PathSeparator}{LatestAzModulePath}";
+                FunctionModulePath = $"{managedDependenciesPath}{Path.PathSeparator}";
```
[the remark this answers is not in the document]
Francisco-Gamino: I agree. Fixed.

src/RequestProcessor.cs
Francisco-Gamino: Fixed.

protobuf/src/proto/FunctionRpc.proto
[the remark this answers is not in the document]
Francisco-Gamino: I will follow-up with @pragnagopa to figure out how to handle this.

src/RequestProcessor.cs
```diff
@@ -86,15 +104,38 @@ internal StreamingMessage ProcessFunctionLoadRequest(StreamingMessage request)
...
+                status.Exception = _unrecoverableFunctionLoadException.ToRpcException();
+                return response;
+            }
+
             try
             {
+                // If functionLoadRequest.ManagedDependencyEnabled is true,
+                // process the function app dependencies defined in functionAppRoot\requirements.psd1.
+                // These dependencies are installed via Save-Module once PowerShell has been initialized.
+                if (functionLoadRequest.ManagedDependencyEnabled && !DependencyManager.FunctionAppDependenciesInstalled)
```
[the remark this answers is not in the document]
Francisco-Gamino: Fixed.

src/RequestProcessor.cs
```diff
@@ -86,15 +104,38 @@ internal StreamingMessage ProcessFunctionLoadRequest(StreamingMessage request)
...
+                {
+                    _dependencyManager.SetFunctionAppDependencies(functionLoadRequest);
+
+                    if (DependencyManager.Dependencies?.Count > 0)
+                    {
+                        response.FunctionLoadResponse.IsDependencyDownloaded = true;
```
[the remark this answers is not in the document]
Francisco-Gamino: This is to let the Host (runtime) know that we have installed the managed dependencies. I have moved this to after running _powershellPool.Initialize(request.RequestId, InstallManagedDependencyModule);

src/RequestProcessor.cs
```diff
@@ -86,15 +104,38 @@ internal StreamingMessage ProcessFunctionLoadRequest(StreamingMessage request)
...
+                }
+
                 // Ideally, the initialization should happen when processing 'WorkerInitRequest', however, the 'WorkerInitRequest'
                 // message doesn't provide the file path of the FunctionApp. That information is not available until the first
                 // 'FunctionLoadRequest' comes in. Therefore, we run initialization here.
                 if (!_isFunctionAppInitialized)
                 {
-                    FunctionLoader.SetupWellKnownPaths(functionLoadRequest);
-                    _powershellPool.Initialize(request.RequestId);
+                    FunctionLoader.SetupWellKnownPaths(functionLoadRequest, DependencyManager.DependenciesPath);
```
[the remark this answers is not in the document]
Francisco-Gamino: Good point. Fixed.

src/RequestProcessor.cs
```diff
@@ -86,15 +104,38 @@ internal StreamingMessage ProcessFunctionLoadRequest(StreamingMessage request)
...
+                    _powershellPool.Initialize(request.RequestId, InstallManagedDependencyModule);
```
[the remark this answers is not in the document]
Francisco-Gamino: Fixed.

src/RequestProcessor.cs
```diff
@@ -6,29 +6,47 @@
...
 
+        // Central repository for acquiring PowerShell modules.
+        private const string Repository = "PSGallery";
+
+        // Az module name
+        private const string AzModuleName = "Az";
+
+        // Holds the exception object if an error is encountered while
+        // initializing PowerShell or downloading the function app dependencies.
+        internal Exception _unrecoverableFunctionLoadException { get; set; }
```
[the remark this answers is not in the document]
Francisco-Gamino: Making field private. The code would look like this:

// Holds the exception if an issue is encountered while installing the function app dependencies
// or while initializing PowerShell.
private Exception UnrecoverableFunctionLoadException { get; set; }

src/RequestProcessor.cs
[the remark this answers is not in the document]
Francisco-Gamino: I have move all this code to the DependencyManager. Now, the delegate calls the InstallManagedDependencyModule in the DependencyManager.

src/RequestProcessor.cs
Francisco-Gamino: Thank you for this suggestion --this makes perfect sense. I have updated the code, so now we append the inner exception message (with what actual failed) to the outer exception. The error message will look like this

Fail to install function app dependencies. Error: '{0}'. The exception message will be created in the delegate that calls DependencyManager.InstallManagedDependencyModule.

src/RequestProcessor.cs
```diff
@@ -251,11 +304,13 @@ private void BindOutputFromResult(InvocationResponse response, AzFunctionInfo fu
...
                         object outValue = results[outBindingName];
-                        object transformedValue = Utils.TransformOutBindingValueAsNeeded(outBindingName, bindingInfo, outValue);
+                        object transformedValue =
+                            Utils.TransformOutBindingValueAsNeeded(outBindingName, bindingInfo, outValue);
                         TypedData dataToUse = transformedValue.ToTypedData();
 
                         // if one of the bindings is '$return' we need to set the ReturnValue
-                        if(string.Equals(outBindingName, AzFunctionInfo.DollarReturn, StringComparison.OrdinalIgnoreCase))
+                        if (string.Equals(outBindingName, AzFunctionInfo.DollarReturn,
+                            StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
Francisco-Gamino: Fixed.

src/DependencyManagement/DependencyManagementUtils.cs
Francisco-Gamino: Updated this method with your latest code.

src/RequestProcessor.cs
Francisco-Gamino: @SatishRanjan : Moving this code to the DependencyManager. Also, this code will be tested by a couple of end to end scenarios I will add.

PR #176, opened 2019-03-27
Removing supportedManagedDependencies.json as the CLI will directly query the PSGallery.

Francisco-Gamino:
Removing supportedManagedDependencies.json as the CLI will directly query the PSGallery.


Francisco-Gamino:
Thanks @TylerLeonhardt and @daxian-dbw !

PR #192, opened 2019-04-17
[Do not merged]Buffered Writes to rpc channel and async everywhere

pragnagopa:
This is an experimental PR for through put issue.  Current implementation follows request/reply pattern for every message sent from the host. 
A grpc write task i.e sending message from host to language worker only completes when the language worker reads the message.  Current implementation of the worker forces host to wait on the grpc response for any request. This specificically impacts invocation requests.


pragnagopa:
@daxian-dbw  - Let's sync up offiline

PR #193, opened 2019-04-20
Managed dependency download on background thread

SatishRanjan:
This PR has the changes to download managed dependency on a background thread and show the managed dependency download in progress message on the azure portal and on the CLI console when a function invocation is triggered while managed dependency download is in-progress

### DIFF
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
     using System.Management.Automation;
     using System.Management.Automation.Language;
     using System.Management.Automation.Runspaces;
+    using System.Threading.Tasks;
 
     internal class DependencyManager
     {
@@ -31,6 +32,9 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         internal static string DependenciesPath { get; private set; }
 
         internal Exception DependencyError { get => _dependencyError; }
+
+        //The dependency download task
+        internal Task DependencyDownloadTask { get => _dependencyDownloadTask; }
 
         // Az module name.
         private const string AzModuleName = "Az";
@@ -61,12 +65,26 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         // If we do, we use it to clean up the module destination path.
         private bool _shouldUpdateFunctionAppDependencies;
 
+        private volatile Task _dependencyDownloadTask;
+
         internal DependencyManager()
         {
             Dependencies = new List<DependencyInfo>();
         }
 
-        internal void ProcessDependencies(
+        internal void ProcessDependencyDownload(MessagingStream msgStream, StreamingMessage request)
+        {
+            if (request.FunctionLoadRequest.ManagedDependencyEnabled)
+            {
+                //Start dependency download on a separate thread
+                _dependencyDownloadTask = Task.Run(() => ProcessDependencies(msgStream, request)).ContinueWith((task) =>
+                {
+                    _dependencyDownloadTask = null;
+                });
+            }
+        }
+
+        private void ProcessDependencies(
             MessagingStream msgStream,
             StreamingMessage request)
         {
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -20,6 +20,8 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
     using Microsoft.Azure.Functions.PowerShellWorker.Messaging;
     using System.Management.Automation;
     using System.Management.Automation.Language;
+    using System.Management.Automation.Runspaces;
+    using System.Threading;
 
     internal class DependencyManager
     {
@@ -28,6 +30,8 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
 
         // This is the location where the dependent modules will be installed.
         internal static string DependenciesPath { get; private set; }
+
+        internal Exception DependencyError { get => _dependencyError; }
 
         // Az module name.
         private const string AzModuleName = "Az";
@@ -52,6 +56,8 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         // Managed Dependencies folder name.
         private const string ManagedDependenciesFolderName = "ManagedDependencies";
 
+        private volatile Exception _dependencyError;
+
         // This flag is used to figure out if we need to install/reinstall all the function app dependencies.
         // If we do, we use it to clean up the module destination path.
         private bool _shouldUpdateFunctionAppDependencies;
@@ -63,39 +69,44 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
 
         internal void ProcessDependencies(
             MessagingStream msgStream,
-            PowerShellManagerPool powershellPool,
-            StreamingMessage request,
-            FunctionLoader functionLoader)
+            StreamingMessage request)
         {
-            PowerShellManager psManager = null;
-            StreamingMessage response = new StreamingMessage { RequestId = request.RequestId };
-            var status = new StatusResult { Status = StatusResult.Types.Status.Success };
-
             try
             {
+                _dependencyError = null;
                 var functionLoadRequest = request.FunctionLoadRequest;
-                // If 'ManagedDependencyEnabled' is true, process the function app dependencies as defined in FunctionAppRoot\requirements.psd1.
-                // These dependencies will be installed via 'Save-Module' when the first PowerShellManager instance is being created.
                 if (functionLoadRequest.ManagedDependencyEnabled)
                 {
                     Initialize(functionLoadRequest);
                 }
 
-                psManager = powershellPool.CheckoutIdleWorker(request);
-                InstallFunctionAppDependencies(psManager.PowerShellInstance, psManager.Logger);
-                response.FunctionLoadResponse.IsDependencyDownloaded = true;
-                functionLoader.LoadFunction(functionLoadRequest);
+                if (_shouldUpdateFunctionAppDependencies)
+                {
+                    var initialSessionState = InitialSessionState.CreateDefault();
+                    initialSessionState.ThreadOptions = PSThreadOptions.UseCurrentThread;
+                    // Setting the execution policy on macOS and Linux throws an exception so only update it on Windows
+                    if (Platform.IsWindows)
+                    {
+                        initialSessionState.ExecutionPolicy = Microsoft.PowerShell.ExecutionPolicy.Unrestricted;
+                    }
+
+                    using (PowerShell PowerShellInstance = PowerShell.Create(initialSessionState))
+                    {
+                        RequestProcessor.IsDependencyDownloadInProgress = true;
+                        var rpcLogger = new RpcLogger(msgStream);
+                        rpcLogger.SetContext(request.RequestId, null);
+                        InstallFunctionAppDependencies(PowerShellInstance, rpcLogger);
+                        RequestProcessor.IsDependencyDownloadInProgress = false;
+                    }
+                }
             }
             catch (Exception e)
             {
-                status.Status = StatusResult.Types.Status.Failure;
-                status.Exception = e.ToRpcException();
+                _dependencyError = e;
             }
             finally
             {
-                powershellPool.ReclaimUsedWorker(psManager);
-                response.FunctionLoadResponse = new FunctionLoadResponse() { Result = status };
-                msgStream.Write(response);
+                RequestProcessor.IsDependencyDownloadInProgress = false;
             }
         }
 
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -74,30 +74,29 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
             try
             {
                 _dependencyError = null;
-                var functionLoadRequest = request.FunctionLoadRequest;
-                if (functionLoadRequest.ManagedDependencyEnabled)
+                var rpcLogger = new RpcLogger(msgStream);
+                rpcLogger.SetContext(request.RequestId, null);
+                if (!_shouldUpdateFunctionAppDependencies)
                 {
-                    Initialize(functionLoadRequest);
+                    // The function app already has the latest dependencies installed.
+                    rpcLogger.Log(LogLevel.Trace, PowerShellWorkerStrings.LatestFunctionAppDependenciesAlreadyInstalled, isUserLog: true);
+                    return;
                 }
 
-                if (_shouldUpdateFunctionAppDependencies)
+                var initialSessionState = InitialSessionState.CreateDefault();
+                initialSessionState.ThreadOptions = PSThreadOptions.UseCurrentThread;
+                initialSessionState.EnvironmentVariables.Add(new SessionStateVariableEntry("PSModulePath", FunctionLoader.FunctionModulePath, null));
+                // Setting the execution policy on macOS and Linux throws an exception so only update it on Windows
+                if (Platform.IsWindows)
                 {
-                    var initialSessionState = InitialSessionState.CreateDefault();
-                    initialSessionState.ThreadOptions = PSThreadOptions.UseCurrentThread;
-                    // Setting the execution policy on macOS and Linux throws an exception so only update it on Windows
-                    if (Platform.IsWindows)
-                    {
-                        initialSessionState.ExecutionPolicy = Microsoft.PowerShell.ExecutionPolicy.Unrestricted;
-                    }
+                    initialSessionState.ExecutionPolicy = Microsoft.PowerShell.ExecutionPolicy.Unrestricted;
+                }
 
-                    using (PowerShell PowerShellInstance = PowerShell.Create(initialSessionState))
-                    {
-                        RequestProcessor.IsDependencyDownloadInProgress = true;
-                        var rpcLogger = new RpcLogger(msgStream);
-                        rpcLogger.SetContext(request.RequestId, null);
-                        InstallFunctionAppDependencies(PowerShellInstance, rpcLogger);
-                        RequestProcessor.IsDependencyDownloadInProgress = false;
-                    }
+                using (PowerShell PowerShellInstance = PowerShell.Create(initialSessionState))
+                {
+                    RequestProcessor.IsDependencyDownloadInProgress = true;
+                    InstallFunctionAppDependencies(PowerShellInstance, rpcLogger);
+                    RequestProcessor.IsDependencyDownloadInProgress = false;
                 }
             }
             catch (Exception e)
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -31,10 +31,10 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         // This is the location where the dependent modules will be installed.
         internal static string DependenciesPath { get; private set; }
 
-        internal Exception DependencyError { get => _dependencyError; }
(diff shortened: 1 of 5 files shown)


SatishRanjan:
@daxian-dbw - could you please take a look at this PR

src/RequestProcessor.cs
```diff
@@ -299,27 +321,15 @@ internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessa
...
-                _powershellPool.Initialize(request.RequestId, _dependencyManager.InstallFunctionAppDependencies);
-                response.FunctionLoadResponse.IsDependencyDownloaded = true;
-            }
-            else
+            FunctionLoader.SetupWellKnownPaths(request.FunctionLoadRequest);
+            if (functionLoadRequest.ManagedDependencyEnabled)
             {
-                _powershellPool.Initialize(request.RequestId);
+                //Start dependency download on a separate thread
+                _dependencyDownloadTask = Task.Run(() => _dependencyManager.ProcessDependencies(_msgStream, request)).ContinueWith((task) =>
```
daxian-dbw: The task creation and the reference to the task should be move to `DependencyManager`. For example, `DependencyManager.Initialize(...)` creates the task and hold a reference to it within `DependencyManager`. In `ProcessInvocationRequest` we can wait on the task by `_dependencyManager.DownloadTask.Wait()` or `_dependencyManager.WaitDownloadToFinish()`.

src/RequestProcessor.cs
```diff
@@ -299,27 +321,15 @@ internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessa
...
+                {
+                    _dependencyDownloadTask = null;
```
daxian-dbw: Are you doing this only for the task to be GC'ed?
Doing this in the continuation means it's theoretically possible that `_dependencyDownloadTask` could become null right before calling `_dependencyDownloadTask.Wait();` in `ProcessInvocationRequest`.

After you move the task creation code to `DependencyManager`, you can set the field to null in `_dependencyManager.WaitDownloadToFinish()`, after the waiting on the task returns.

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -26,41 +26,41 @@
...
     mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array
+    value   : The object must be serialized into a byte array 
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->
-  <xsd:schema id="root" xmlns="" xmlns:xsd="https://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
-    <xsd:import namespace="https://www.w3.org/XML/1998/namespace" />
+  <xsd:schema id="root" xmlns="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
+    <xsd:import namespace="http://www.w3.org/XML/1998/namespace" />
```
daxian-dbw: Please revert changes in these 2 lines.

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!--
```
daxian-dbw: Please revert all the extra trailing spaces introduced in this file.

src/RequestProcessor.cs
```diff
@@ -209,6 +211,28 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
                 functionInfo = _functionLoader.GetFunctionInfo(request.InvocationRequest.FunctionId);
                 psManager = _powershellPool.CheckoutIdleWorker(request, functionInfo);
 
+                if (_dependencyDownloadTask != null
+                    && ((_dependencyDownloadTask.Status != TaskStatus.Canceled)
+                    || _dependencyDownloadTask.Status != TaskStatus.Faulted
+                    || _dependencyDownloadTask.Status != TaskStatus.RanToCompletion))
+                {
+                    psManager.Logger.Log(LogLevel.Information, PowerShellWorkerStrings.DependencyDownloadInProgress, null, true);
+                    psManager.Logger.Log(LogLevel.Information, PowerShellWorkerStrings.DependencyDownloadInProgress, null);
```
daxian-dbw: Why writing this again into the system log? This message is only for user to see as a UX experience improvement, isn't it?

src/RequestProcessor.cs
```diff
@@ -209,6 +211,28 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
...
+
+                if (_dependencyManager?.DependencyError != null)
+                {
+                    StreamingMessage response = NewStreamingMessageTemplate(request.RequestId,
+                        StreamingMessage.ContentOneofCase.InvocationResponse,
+                        out StatusResult status);
+                    status.Status = StatusResult.Types.Status.Failure;
+                    status.Exception = _dependencyManager.DependencyError.ToRpcException();
+                    response.InvocationResponse.InvocationId = request.InvocationRequest.InvocationId;
+                    return response;
```
daxian-dbw: As far as I understand, returning a error response to host in `FunctionLoadRequest` phase meaning a hard failure -- all subsequent invocation will fail with the same error.
Failure in `InvocationRequest` phase is not considered a hard failure. This means the next invocation comes in, and it will just continue to execute just like the dependency download/check was successful.

src/RequestProcessor.cs
```diff
@@ -234,7 +257,6 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
                 response.InvocationResponse.InvocationId = request.InvocationRequest.InvocationId;
                 status.Status = StatusResult.Types.Status.Failure;
                 status.Exception = e.ToRpcException();
-
```
daxian-dbw: Please add back those two new lines. They make the code more readable.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -17,8 +17,11 @@
 
 namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
 {
+    using Microsoft.Azure.Functions.PowerShellWorker.Messaging;
     using System.Management.Automation;
     using System.Management.Automation.Language;
+    using System.Management.Automation.Runspaces;
+    using System.Threading;
```
daxian-dbw: Please move `using Microsoft.Azure.Functions.PowerShellWorker.Messaging;` and `using System.Threading;` to the outer using directives.
The reason `using System.Management.Automation;` is put here is because that's how you can use `PowerShell` type in the following code without having to reference to it by the full type name. So it's OK to put all `System.Management.Automation.*` namespace directive here, but not the rest.

src/PowerShell/PowerShellManager.cs
```diff
@@ -33,6 +33,11 @@ internal class PowerShellManager
         /// </summary>
         internal ILogger Logger => _logger;
 
+        /// <summary>
+        /// Gets the associated PowerShell instance
+        /// </summary>
+        internal PowerShell PowerShellInstance => _pwsh;
```
daxian-dbw: This is not used anywhere, so please remove it.

src/PowerShell/PowerShellManagerPool.cs
```diff
@@ -50,14 +50,14 @@ internal PowerShellManagerPool(MessagingStream msgStream)
         /// Initialize the pool and populate it with PowerShellManager instances.
         /// We instantiate PowerShellManager instances in a lazy way, starting from size 1 and increase the number of workers as needed.
         /// </summary>
-        internal void Initialize(string requestId, Action<PowerShell, ILogger> initAction = null)
+        internal void Initialize(string requestId)
```
daxian-dbw: This method can be removed if not needed anymore.

src/PowerShell/PowerShellManagerPool.cs
```diff
@@ -98,7 +98,7 @@ internal PowerShellManager CheckoutIdleWorker(StreamingMessage request, AzFuncti
                 }
             }
 
-            // Register the function with the Runspace before returning the idle PowerShellManager.
+            // Register the function with the Runspace PowerShellManager.
```
daxian-dbw: I think the original comment is more clear. Can you please revert this change?

src/RequestProcessor.cs
daxian-dbw: Also, a similar issue -- we have to create a `Runspace` and evaluate `profile.ps1` in `FunctionLoad` phase, so that we get a hard failure if `profile.ps1` fails to run.

src/RequestProcessor.cs
```diff
@@ -209,6 +211,28 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
...
                 psManager = _powershellPool.CheckoutIdleWorker(request, functionInfo);
 
+                if (_dependencyDownloadTask != null
+                    && ((_dependencyDownloadTask.Status != TaskStatus.Canceled)
+                    || _dependencyDownloadTask.Status != TaskStatus.Faulted
+                    || _dependencyDownloadTask.Status != TaskStatus.RanToCompletion))
+                {
+                    psManager.Logger.Log(LogLevel.Information, PowerShellWorkerStrings.DependencyDownloadInProgress, null, true);
+                    psManager.Logger.Log(LogLevel.Information, PowerShellWorkerStrings.DependencyDownloadInProgress, null);
+                    _dependencyDownloadTask.Wait();
```
pragnagopa: Use async await instead.

pragnagopa:
@daxian-dbw - I will work with @fabiocav to enable PR builds.

src/RequestProcessor.cs
pragnagopa: I think better approach would be to return a Rpc messag and block until dependencies are downloaded as invocation will not succeed before download completes.

src/RequestProcessor.cs
pragnagopa: Yes. FunctioLoad failure is preserved and is replayed on function invocation. Scope of an invocation request failure is only for that invocation

src/RequestProcessor.cs
pragnagopa: We should add another state to the [StatusResult](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/dev/src/proto/FunctionRpc.proto#L112) `Loading`  to allow host to handle intermediate state where function is still loading but not ready for invocation

daxian-dbw: That sounds like a much better long term solution

src/RequestProcessor.cs
SatishRanjan: If DependencyError download error occurs then for all invocation will receive this error and function execution wouldn't proceed

daxian-dbw: You are basically moving the logic that's supposed to be handled by the host to the worker.

See my comment https://github.com/Azure/azure-functions-powershell-worker/pull/193#discussion_r277522973
I believe `profile.ps1` needs to be evaluated during function load phase, because a bad `profile.ps1` is a hard failure. But `profile.ps1` cannot be invoked until after the dependency download.

src/RequestProcessor.cs
SatishRanjan: The PM requirement is to show the message on the azure portal when a function is invoked through portal while managed dependency download is in-progress

src/resources/PowerShellWorkerStrings.resx
SatishRanjan: These spaces are auto formatted

src/RequestProcessor.cs
SatishRanjan: System log is to appear on CLI console

src/RequestProcessor.cs
SatishRanjan: Since dependency download happens on a separate instance of the PowerShell and doesn't use the runspace of powershell pool  hence I've added back the __powershellPool.Initialize(request.RequestId); as part of FunctionLoad

src/PowerShell/PowerShellManagerPool.cs
SatishRanjan: Removed

pragnagopa:
Will enable builds ASAP. Will update this PR soon.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -51,15 +59,83 @@ internal class DependencyManager
...
+        internal void ProcessDependencyDownload(MessagingStream msgStream, StreamingMessage request)
+        {
+            if (request.FunctionLoadRequest.ManagedDependencyEnabled)
+            {
+                var rpcLogger = new RpcLogger(msgStream);
+                rpcLogger.SetContext(request.RequestId, null);
+                if (!_shouldUpdateFunctionAppDependencies)
+                {
+                    // The function app already has the latest dependencies installed.
+                    rpcLogger.Log(LogLevel.Information, PowerShellWorkerStrings.LatestFunctionAppDependenciesAlreadyInstalled, isUserLog: true);
```
daxian-dbw: This message is not accurate. It's possible the `requirements.ps1` only contains an empty hashable -- that's the way for a user to opt out the managed dependency if they don't depend on Az module at all.
Please take a look at how log messages are handled in `InstallFunctionAppDependencies` when `if (Dependencies.Count == 0)` and `if (!_shouldUpdateFunctionAppDependencies)`, and it should be handled the same way.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -51,15 +59,83 @@ internal class DependencyManager
         // Managed Dependencies folder name.
         private const string ManagedDependenciesFolderName = "ManagedDependencies";
 
+        private volatile Exception _dependencyError;
+
         // This flag is used to figure out if we need to install/reinstall all the function app dependencies.
         // If we do, we use it to clean up the module destination path.
         private bool _shouldUpdateFunctionAppDependencies;
 
+        private volatile Task _dependencyDownloadTask;
```
daxian-dbw: `volatile` is not needed for both `_dependencyError` and `_dependencyDownloadTask`. They are not being modified on two threads at the same time.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -51,15 +59,83 @@ internal class DependencyManager
...
+        {
+            if (_dependencyDownloadTask != null)
+            {
+                _dependencyDownloadTask.Wait();
+                _dependencyDownloadTask = null;
+            }
+        }
+
+        private void ProcessDependencies(
+            MessagingStream msgStream,
```
daxian-dbw: Parameter not used.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -51,15 +59,83 @@ internal class DependencyManager
...
+            StreamingMessage request,
```
daxian-dbw: Parameter not used.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -28,6 +31,11 @@ internal class DependencyManager
         // This is the location where the dependent modules will be installed.
         internal static string DependenciesPath { get; private set; }
 
+        internal Exception DependencyError { get => _dependencyError; }
+
+        //The dependency download task
+        internal Task DependencyDownloadTask { get => _dependencyDownloadTask; }
```
daxian-dbw:
```suggestion
        internal Task DependencyDownloadTask => _dependencyDownloadTask;
```

src/DependencyManagement/DependencyManager.cs
```diff
@@ -28,6 +31,11 @@ internal class DependencyManager
         // This is the location where the dependent modules will be installed.
         internal static string DependenciesPath { get; private set; }
 
+        internal Exception DependencyError { get => _dependencyError; }
```
daxian-dbw:
```suggestion
        internal Exception DependencyError => _dependencyError;
```

fabiocav:
Reopening PR to test CI integration

src/RequestProcessor.cs
```diff
@@ -206,9 +205,33 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
...
             {
+                if (_dependencyManager.DependencyDownloadTask != null
+                    && ((_dependencyManager.DependencyDownloadTask.Status != TaskStatus.Canceled)
+                    || _dependencyManager.DependencyDownloadTask.Status != TaskStatus.Faulted
+                    || _dependencyManager.DependencyDownloadTask.Status != TaskStatus.RanToCompletion))
+                {
+                    var rpcLogger = new RpcLogger(_msgStream);
+                    rpcLogger.SetContext(request.RequestId, request.InvocationRequest?.InvocationId);
+                    rpcLogger.Log(LogLevel.Information, PowerShellWorkerStrings.DependencyDownloadInProgress, null, true);
+                    rpcLogger.Log(LogLevel.Information, PowerShellWorkerStrings.DependencyDownloadInProgress, null);
```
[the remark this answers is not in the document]
daxian-dbw: #closed.

src/RequestProcessor.cs
```diff
@@ -206,9 +205,33 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
...
+                    _dependencyManager.DependencyDownloadTask.Wait();
```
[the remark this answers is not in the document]
daxian-dbw: #closed.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -60,6 +74,74 @@ internal DependencyManager()
...
+        internal void ProcessDependencyDownload(MessagingStream msgStream, StreamingMessage request)
+        {
+            if (request.FunctionLoadRequest.ManagedDependencyEnabled)
+            {
+                var rpcLogger = new RpcLogger(msgStream);
+                rpcLogger.SetContext(request.RequestId, null);
+                if (Dependencies.Count == 0)
+                {
+                    // If there are no dependencies to install, log and return.
+                    rpcLogger.Log(LogLevel.Trace, PowerShellWorkerStrings.FunctionAppDoesNotHaveDependentModulesToInstall, isUserLog: true);
```
daxian-dbw: Could it be possible that the portal looks like freezing currently without any message only because we are using the `Trace` level logging? Maybe if you change the current logging to `Information` level and the message will show up even without this change?

src/RequestProcessor.cs
```diff
@@ -299,28 +318,14 @@ internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessa
         private void InitializeForFunctionApp(StreamingMessage request, StreamingMessage response)
```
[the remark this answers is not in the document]
SatishRanjan: I'll perform the protobuf cleanup in the next PR

src/DependencyManagement/DependencyManager.cs
SatishRanjan: I tried changing the trace level but it doesn't show up on azure portal unless it's exception.

src/RequestProcessor.cs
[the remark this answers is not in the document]
SatishRanjan: Cleaned up the StreamingMessage response parameter

src/RequestProcessor.cs
```diff
@@ -225,7 +247,6 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
             catch (Exception e)
             {
                 _powershellPool.ReclaimUsedWorker(psManager);
-
```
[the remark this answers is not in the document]
SatishRanjan: Added the new line

src/RequestProcessor.cs
```diff
@@ -234,7 +255,6 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
                 response.InvocationResponse.InvocationId = request.InvocationRequest.InvocationId;
                 status.Status = StatusResult.Types.Status.Failure;
                 status.Exception = e.ToRpcException();
-
```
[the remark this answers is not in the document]
SatishRanjan: added new line

PR #194, opened 2019-04-25
Refactor and clean up based on the dependency management changes

daxian-dbw:
Refactor and clean up the code base based on the dependency management changes.
- Reduce redundant code -- `initialsessionstate` creation and failure response in `FunctionLoad` processing
- Remove the unnecessary `_powershellPool.ReclaimUsedWorker(psManager);` in `ProcessInvocationRequest`.
- Change the line ending from `CRLF` to `LF` in `Utils.cs`.
   - It's easier to review with https://github.com/Azure/azure-functions-powershell-worker/pull/194/files?w=1

### DIFF
--- a/src/Utility/Utils.cs
+++ b/src/Utility/Utils.cs
@@ -6,6 +6,7 @@
 using System;
 using System.IO;
 using System.Management.Automation;
+using System.Management.Automation.Runspaces;
 using System.Text;
 using Microsoft.PowerShell.Commands;
 
@@ -17,6 +18,34 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Utility
         internal static CmdletInfo RemoveModuleCmdletInfo = new CmdletInfo("Remove-Module", typeof(RemoveModuleCommand));
         internal static CmdletInfo GetJobCmdletInfo = new CmdletInfo("Get-Job", typeof(GetJobCommand));
         internal static CmdletInfo RemoveJobCmdletInfo = new CmdletInfo("Remove-Job", typeof(RemoveJobCommand));
+
+        private static InitialSessionState s_initialSessionState = null;
+
+        internal static InitialSessionState GetInitialSessionState()
+        {
+            if (s_initialSessionState == null)
+            {
+                var iss = InitialSessionState.CreateDefault();
+                iss.ThreadOptions = PSThreadOptions.UseCurrentThread;
+                iss.EnvironmentVariables.Add(
+                    new SessionStateVariableEntry(
+                        "PSModulePath",
+                        FunctionLoader.FunctionModulePath,
+                        description: null));
+
+                // Setting the execution policy on macOS and Linux throws an exception so only update it on Windows
+                if(Platform.IsWindows)
+                {
+                    // This sets the execution policy on Windows to Unrestricted which is required to run the user's function scripts on
+                    // Windows client versions. This is needed if a user is testing their function locally with the func CLI.
+                    iss.ExecutionPolicy = Microsoft.PowerShell.ExecutionPolicy.Unrestricted;
+                }
+
+                s_initialSessionState = iss;
+            }
+
+            return s_initialSessionState;
+        }
 
         /// <summary>
         /// Helper method to do additional transformation on the input value based on the type constraints specified in the script.
--- a/src/Utility/Utils.cs
+++ b/src/Utility/Utils.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Management.Automation;
 using System.Management.Automation.Runspaces;
 using System.Text;
+using System.Threading;
 using Microsoft.PowerShell.Commands;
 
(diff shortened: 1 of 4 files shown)


daxian-dbw:
@SatishRanjan and @TylerLeonhardt Can you please take a look at this PR?
This PR does some minor refactoring to reduce redundancy, and change the line ending in `Utils.cs`.

src/Utility/Utils.cs
```diff
@@ -1,102 +1,131 @@
...
+
+                // Setting the execution policy on macOS and Linux throws an exception so only update it on Windows
+                if(Platform.IsWindows)
+                {
+                    // This sets the execution policy on Windows to Unrestricted which is required to run the user's function scripts on
+                    // Windows client versions. This is needed if a user is testing their function locally with the func CLI.
+                    iss.ExecutionPolicy = Microsoft.PowerShell.ExecutionPolicy.Unrestricted;
+                }
+
+                s_initialSessionState = iss;
```
SatishRanjan: Is reusing same instance of InitialSessionState not going to cause issue especially when PowerShell instance pool size is greater than 1, say for example if an instance of PowerShell goes wrong then during Dispose associated session will be disposed as well which will cause other PowerShell instance not function correctly?

src/Utility/Utils.cs
```diff
@@ -1,102 +1,131 @@
...
+    {
+        internal static CmdletInfo ImportModuleCmdletInfo = new CmdletInfo("Import-Module", typeof(ImportModuleCommand));
+        internal static CmdletInfo RemoveModuleCmdletInfo = new CmdletInfo("Remove-Module", typeof(RemoveModuleCommand));
+        internal static CmdletInfo GetJobCmdletInfo = new CmdletInfo("Get-Job", typeof(GetJobCommand));
+        internal static CmdletInfo RemoveJobCmdletInfo = new CmdletInfo("Remove-Job", typeof(RemoveJobCommand));
+
+        private static InitialSessionState s_initialSessionState = null;
+
+        internal static InitialSessionState GetInitialSessionState()
+        {
```
SatishRanjan: This can be changed to Lazy and expose InitialSessionState as property and make GetInitialSessionState() as private to just create an instance of InitialSessionState like below:
 private static Lazy<InitialSessionState> s_initialSessionState = new Lazy<InitialSessionState>(GetInitialSessionState);
        internal static InitialSessionState InitialSessionStateInstance => s_initialSessionState.Value;

        private static InitialSessionState GetInitialSessionState()
        {
            var iss = InitialSessionState.CreateDefault();
            iss.ThreadOptions = PSThreadOptions.UseCurrentThread;
            iss.EnvironmentVariables.Add(
                new SessionStateVariableEntry(
                    "PSModulePath",
                    FunctionLoader.FunctionModulePath,
                    description: null));

            // Setting the execution policy on macOS and Linux throws an exception so only update it on Windows
            if (Platform.IsWindows)
            {
                // This sets the execution policy on Windows to Unrestricted which is required to run the user's function scripts on
                // Windows client versions. This is needed if a user is testing their function locally with the func CLI.
                iss.ExecutionPolicy = Microsoft.PowerShell.ExecutionPolicy.Unrestricted;
            }

            return iss;
        }

src/Utility/Utils.cs
daxian-dbw: I'm glad you asked this question!

A cloned copy of the passed in `initialsessionstate` will be actually used for the Runspace creation, see https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/hostifaces/ConnectionBase.cs#L78
So it's safe to reuse the same `initialsessionstate` to create multiple Runspaces.

src/Utility/Utils.cs
daxian-dbw: Sure, I can make it a `Lazy<T>` property.

src/Utility/Utils.cs
```diff
@@ -1,102 +1,128 @@
...
+namespace Microsoft.Azure.Functions.PowerShellWorker.Utility
+{
+    internal class Utils
+    {
+        internal static CmdletInfo ImportModuleCmdletInfo = new CmdletInfo("Import-Module", typeof(ImportModuleCommand));
+        internal static CmdletInfo RemoveModuleCmdletInfo = new CmdletInfo("Remove-Module", typeof(RemoveModuleCommand));
+        internal static CmdletInfo GetJobCmdletInfo = new CmdletInfo("Get-Job", typeof(GetJobCommand));
+        internal static CmdletInfo RemoveJobCmdletInfo = new CmdletInfo("Remove-Job", typeof(RemoveJobCommand));
+
+        internal static Lazy<InitialSessionState> SingletonISS
```
SatishRanjan: Please make this readonly otherwise it would be prone to get modified which is not the desired behavior

daxian-dbw: Already addressed in my last commit.

PR #214, opened 2019-05-08
Avoid invoking `run.ps1` and port helper module to C# to improve throughput performance

### LINKED ISSUE #127: Investigate throughput result when targeting a single powershell worker
The throughput tests targeting a single powershell worker doesn't show cheering results:
- When targeting a PowerShell worker deployed in a dedicated Azure sandbox, we see around 30 RPS regardless the size of `PowerShellManager` pool.
  The average response time shown in WCAT is between 3300ms to 3500ms.
- When targeting a PowerShell worker running locally on my dev machine, we see around 46 RPS with pool size being 5 (better than size being 1).
  The average response time shown in [LOCUS](https://locust.io/) keeps increasing to over 2000 ms when the velocity of the in-coming request reaches about 70/s.

Based on the data collected from both environments, the average time spent in PS worker for each invocation request is minimal -- less than 100ms for the Azure sandbox setup, and less than 50ms for the local setup.

I'm suspecting the overhead is in the Host. I have discussed with @pragnagopa and we believe it would be the best for the Host to be able to measure the average time for an invocation request to be processed by a worker (time from sending the invocation request to the worker to receiving the invocation response from the worker).

-----

**Local setup data and analysis:**

https://github.com/Azure/azure-functions-powershell-worker/pull/123#issue-242419384

-----

**Azure sandbox data and analysis:**

This throughput test was done targeting PS worker in a App Service Plan sandbox:

The PS worker used in this setup includes some instrumentation code to collect data about the time spent in PS worker, including:
-	total number of request received from the gPRC channel [Request]
-	total number of invocation request processed (a task is created for the invocation request, but not necessarily scheduled) [InvRequest]
-	total number of invocation request finished processing [InvDone]
-	total number of invocation response written out to host [Response]
-	average time for retrieving an invocation request [AveGetReqTime] = _sumReadRequestTime/InvRequest
-	average time for processing an invocation [AveInvTime] = _sumInvTime/InvDone
-	average time for fetching a PSManager instance from the pool [AveFetchFromPoolTime] = _sumFetchFromPoolTime/InvRequest
-	average time for writing a invocation response to host [AveResponseTime] = _sumResponseTime/Response

I used WCAT with 100 users and 3 min duration. As you can see in the log below, the average response time is around 3500 ms

temp

However, the data collection from the PS worker indicates there isn’t much time spent in the worker. The data collected are listed blow (also attached as in `pool-1-100user-3min.txt`).
As you can see, the average time for all operations in PS worker is decreasing gradually. The average invocation time and average fetch-from-pool time are mostly constant (23ms and 1.5ms).

Take the highlighted fraction as an example, PS worker processes 500 invocation requests during that time, and the average time for processing each invocation request during that time is 87.998ms.
`$totalTimeSpentAfter4602 = 34.7941304347826 * 4600 + 22.6766688410524 * 4599 + 1.68565217391304 * 4600 + 36.4443236189648 * 4598  ## (result: 439668ms)`
`$totalTimeSpentAfter5102 = 34.9847058823529 * 5100 + 22.8405569719553 * 5099 + 1.57372549019608 * 5100 + 36.6294625343272 * 5098 ## (result: 489649ms)`
`$aveTimeFor500Inv = ($totalTimeSpentAfter5102 - $totalTimeSpentAfter4602) / 500 ## (result: 99.962ms)`

So it looks to me most of the time was spent in the host, which matches the observation I got when running throughput tests locally on my dev machine.
This is really puzzling to me, and any help to understand it would be really appreciated.

```
Request,InvRequest,InvDone,Response,AveGetReqTime,AveInvTime,AveFetchFromPoolTime,AveResponseTime
102,100,99,69,27.12,42.9090909090909,18.19,39.8115942028985
202,200,199,192,38.27,32.7336683417085,9.695,40.7760416666667
302,300,299,296,34.9566666666667,29.61872909699,6.95,36.5506756756757
402,400,399,397,33.945,27.4260651629073,5.2125,35.0856423173804
502,500,499,495,31.884,27.503006012024,6.652,35.4888888888889
602,600,599,597,32.095,26.7345575959933,5.59,35.4505862646566
702,700,699,698,32.1071428571429,25.931330472103,4.84428571428571,35.0716332378223
802,800,799,798,42.19625,25.3128911138924,4.23875,44.859649122807
902,900,899,898,41.9455555555556,24.9866518353726,3.76777777777778,44.3507795100223
1002,1000,999,998,40.866,24.6136136136136,3.396,43.060120240481
1102,1100,1099,1098,39.8018181818182,24.3721565059145,3.11545454545455,41.8624772313297
1202,1200,1199,1198,39.9525,24.0892410341952,2.85583333333333,41.8622704507513
1302,1300,1299,1298,39.1776923076923,23.8306389530408,2.64923076923077,40.9722650231125
1402,1400,1399,1398,38.6421428571429,23.6182987848463,2.46357142857143,40.3354792560801
1502,1500,1499,1498,38.1406666666667,23.5350233488993,2.29933333333333,39.714953271028
1602,1600,1599,1598,37.72125,23.4521575984991,2.155625,39.2340425531915
1702,1700,1699,1698,37.3723529411765,23.3778693349029,2.03470588235294,38.8186101295642
1802,1800,1799,1798,36.9511111111111,23.2051139521957,1.92166666666667,38.3286985539488
1902,1900,1899,1897,36.6357894736842,23.1337546076883,1.82526315789474,37.9615181866104
2002,2000,1999,1998,36.3505,23.0525262631316,1.734,37.6151151151151
2102,2100,2099,2093,36.1504761904762,23.2853739876131,1.8647619047619,37.6684185379838
2202,2200,2199,2198,36.2,23.2201000454752,1.80818181818182,37.5973612374886
2302,2300,2299,2298,36.0017391304348,23.1239669421488,1.7295652173913,37.3424717145344
2402,2400,2399,2385,35.96375,23.0070862859525,1.68541666666667,37.5014675052411
2502,2500,2499,2497,35.3568,23.046418567427,2.1656,37.2378854625551
2602,2600,2599,2598,35.0684615384615,23.042323970758,2.17538461538462,36.9949961508853
2702,2700,2699,2698,34.9351851851852,23.0070396443127,2.11111111111111,36.8150481838399
2802,2800,2799,2798,34.8078571428571,22.9521257591997,2.03571428571429,36.6361686919228
2902,2900,2899,2898,34.6675862068966,22.8868575370818,1.96965517241379,36.4482401656315
3002,3000,2999,2998,34.6786666666667,22.8589529843281,1.90566666666667,36.4149432955304
3102,3100,3099,3096,34.4174193548387,22.8076798967409,1.89387096774194,36.187984496124
3202,3200,3199,3198,34.7071875,22.7596123788684,1.8346875,36.3977485928705
3302,3300,3299,3298,34.5327272727273,22.6968778417702,1.78272727272727,36.1770770163736
3402,3400,3399,3390,34.4988235294118,22.7069726390115,1.74705882352941,36.1693215339233
3502,3500,3499,3498,34.4234285714286,22.6684767076308,1.75057142857143,36.0711835334477
3602,3600,3599,3598,34.2975,22.6368435676577,1.70194444444444,35.9043913285158
3702,3700,3699,3698,34.1891891891892,22.5844822925115,1.66405405405405,35.7644672796106
3802,3800,3799,3795,33.8494736842105,22.6459594630166,1.92868421052632,35.7191040843215
3902,3900,3899,3898,34.5941025641026,22.6358040523211,1.89717948717949,36.4256028732683
4002,4000,3999,3998,34.5555,22.6006501625406,1.851,36.3554277138569
4102,4100,4099,4097,34.4546341463415,22.6216150280556,1.88341463414634,36.3099829143276
4202,4200,4199,4198,34.49,22.5999047392236,1.83857142857143,36.3120533587423
4302,4300,4299,4298,34.5125581395349,22.5887415678065,1.79697674418605,36.3008375988832
4402,4400,4398,4397,34.5706818181818,22.6339245111414,1.75613636363636,36.2972481237207
4502,4500,4499,4498,34.5691111111111,22.6463658590798,1.71711111111111,36.2865718096932
4602,4600,4599,4598,34.7941304347826,22.6766688410524,1.68565217391304,36.4443236189648
4702,4700,4699,4698,35.1448936170213,22.7961268354969,1.65106382978723,36.8288633461047
4802,4800,4799,4796,35.1420833333333,22.8524692644301,1.654375,36.8302752293578
4902,4900,4899,4898,35.15,22.8591549295775,1.62877551020408,36.8380971825235
5002,5000,4999,4998,35.1224,22.8499699939988,1.5962,36.7851140456183
5102,5100,5099,5098,34.9847058823529,22.8405569719553,1.57372549019608,36.6294625343272
```

daxian-dbw:
Summary
-----------
Fix #127
Note that, the NO.1 change below is a big design change, and I would like your help to think about all scenarios that might not work with this change.
For example, @TylerLeonhardt had the great concern whether this will continue to work with local debugging. I tested it and it still works.

Major changes
----------------

1. Deploy the Azure Function scripts (e.g. `run.ps1`) to `Runspace` as actual PowerShell functions, so as to avoid hitting the file system for every invocation when using `AddCommand(<path-to-ps1-file>).Invoke()`.
    - In the sandbox environment, the FunctionApp files are on the user's storage account, and thus file system access is expensive. Invoking an in-memory PS function is 8x faster than invoking a `.ps1` file in that environment.
    - Note that, the PS functions we deploy will be **constant functions** -- they cannot be modified or removed. So we don't need to worry about the function being changed on the fly.
    - `#Requires` only works when invoking the script file directly, so we don't do this deployment if `#Requires` is used in the function script and remain invoking the `.ps1` file directly in that case.
    - After turning a `.ps1` file to a PS function, script variables like `$script:xxx` used in the script will be made global variables as there is no script scope anymore. So we have to clean up global variables after every invocation.
    - Currently, we call `ResetRunspaceState()` to clean up variable table. Unfortunately, it does more than that -- it also reset the `Debugger`, create new `EventManager`, reset the `CurrentPath`. Resetting `Debugger` causes a problem to our local debugging experience -- **breakpoints set during the first invocation debugging cannot be hit in the subsequent invocations**. All we need is to clean up the global variables newly added by an invocation, so I replace `ResetRunspaceState()` with simply code that should have the similar functionality.

2. Rewrite the helper module in C#, for two reasons: (1) improve performance (2) we will need to have C# cmdlets when getting to durable function implementation anyways. This change has the following impact:
    - The helper module cannot be loaded by a version of pwsh that is lower than the PowerShell NuGet packages we reference in the worker, because of the specific version of `Newtonsoft.Json` dependency. I talked to @joeyaiello about this, and we believe it's OK to do this, as we don't think it's useful to load this helper module outside of the worker in the first place. This change won't help us to have the auto-completion UX for the helper commands in VSCode, but it's a separate issue.
    - We cannot use `Pester` to test the helper module, so I move all pester tests to xUnit.
    - We cannot do the markdown help doc check anymore (cannot load the helper module unless download the right version of `pwsh`)

Next steps
------------

- [ ] Collect feedback and concerns for scenarios and do verification
   - [x] Local debugging
      - DONE. Local debugging works as expected, and the breakpoint issue highlighted above is also resolved.
   - [x] Special variables like `$PSScriptRoot` and `$PSCommandPath`
      - DONE. They work as expected. `$PSScriptRoot` points to the script file's parent directory. `PSCommandPath` points to the script file itself.
      - xUnit test added for this scenario.
   - [x] Error should continue to report the correct script stack trace and line/column offset. (Thanks @JustinGrote!)
      - DONE. The error record works as expected. Script stack trace and line/column offset are the same as running the `.ps1` directly.
   - [ ] More ... pending on feedback
- [x] Collect the throughput data with the final changes

Throughput measurements with the final code change
---------

I measured for **20 times** with a dedicated VM (1 CPU core, 1.7 GB memory).

### Testing environment

I used a **fresh sandbox process** for every measurement with the following steps:

1. Run `net stop dwassvc` to stop the service
2. Run `net start dwassvc` to start the service
3. Wait for CPU to come down
4. Run `irm <http-trigger-uri>` for 3,4 times to pass the cold start.
5. Start throughput test using `WCAT Runner`, with this setting: **100 Users, 1 minute Duration**
    - `WCAT` first uses 30 seconds to warm up the service, then collects the **average Response Per Second (RPS)** value every 10 seconds for 1 minute

### Testing FunctionApp

An HTTP function that returns a OK response for every invocation.

```powershell
param($req, $TriggerMetadata)
Push-OutputBinding -Name res -Value ([HttpResponseContext]@{
  StatusCode = 200
  Body = "Hello"
})
```

#### Overall statistics for the 20 measurements (the last column from the detailed data set table below)

```
PS> 74, 70, 89, 69, 85, 83, 76, 68, 68, 70, 73, 80, 74, 69, 85, 73, 84, 68, 76, 73 | Measure-Object -AllStats

Count             : 20
Average           : 75.35
Sum               : 1507
Maximum           : 89
Minimum           : 68
StandardDeviation : 6.68285315683275
```

According to Dormann 2013 CV-values below 0.05 (5%) indicate very high precision of the data, values above 0.2 (20%) low precision (a rule of thumb).
The CV value of my measurement data set is 0.0887 (8.87%), so the data are **relatively precise based on the rule of thumb**.

| Count | Sum | Maximum | Minimum | Mean | Std. Deviation | coefficient of variation (CV) |
| -------| ------| -----------| ------------| -------| ----------------| ------------------------------|
| 20 | 1507 | 89 | 68 | 75.35 | 6.683 | 0.0887 (Std. Dev. / Mean) |

#### Detailed measurement data set

`WCAT` collects the **average Response Per Second (RPS)** value every 10 seconds for 1 minute

| Id  | 5 sec | 15 sec | 25 sec | 35 sec | 45 sec | 55 sec | 60 sec |
| ----| ----| ----| ----| ----| ----| ----| ----|
| 1 | 77 | 77 | 81 | 82 | 78 | 74 | 74 |
| 2 | 73 | 73 | 73 | 72 | 72 | 70 | 70 |
| 3 | 45 | 87 | 88 | 88 | 87 | 88 | 89 |
| 4 | 60 | 64 | 69 | 70 | 69 | -- | 69 |
| 5 | 69 | 79 | 83 | 84 | 85 | 86 | 85 |
| 6 | 72 | 84 | 87 | 87 | 86 | 85 | 83 |
| 7 | 79 | 75 | 81 | 82 | 79 | 76 | 76 |
| 8 | 67 | 66 | 67 | 65 | 69 | 68 | 68 |
| 9 | 68 | 58 | 62 | 67 | 69 | 68 | 68 |
| 10 | 68 | 75 | 73 | 72 | 71 | 70 | 70 |
| 11 | 56 | 73 | 71 | 71 | 72 | 73 | 73 |
| 12 | 78 | 84 | 85 | 85 | 81 | 80 | 80 |
| 13 | 73 | 77 | 81 | 81 | 76 | 74 | 74 |
| 14 | 54 | 66 | 69 | 68 | 69 | 68 | 69 |
| 15 | 77 | 84 | 85 | 87 | 87 | 87 | 85 |
| 16 | 72 | 74 | 73 | 72 | 72 | 72 | 73 |
| 17 | 75 | 72 | 78 | 82 | 84 | -- | 84 |
| 18 | 59 | 61 | 66 | 66 | 69 | 68 | 68 |
| 19 | 84 | 73 | 75 | 78 | 77 | 76 | 76 |
| 20 | 54 | 71 | 71 | 73 | 73 | 72 | 73 |

### Measurement data WITHOUT changes in this PR

Before this PR, the average RPS is 22.3. This PR **achieves over 3 time improvement (75.35 / 22.3 = 3.37)**

| Time | Test | Runtime | RPS | Total requests | Average response time |
| ---| ---| ---| ---| ---| ---|
|2019/05/13 14:34:26 | PS Ping | 2.0.12455.0 | 20 (Threshold:20) | 2468 | 4941.86|
|2019/05/12 14:35:25 | PS Ping | 2.0.12455.0 | 21 (Threshold:20) | 2543 | 4792.75|
|2019/05/11 14:33:38 | PS Ping | 2.0.12455.0 | 23 (Threshold:20) | 2878 | 4219.34|
|2019/05/10 14:33:26 | PS Ping | 2.0.12453.0 | 24 (Threshold:20) | 2889 | 4201.02|
|2019/05/09 14:34:31 | PS Ping | 2.0.12451.0 | 21 (Threshold:20) | 2530 | 4797.76|
|2019/05/08 14:33:00 | PS Ping | 2.0.12451.0 | 22 (Threshold:20) | 2756 | 4405.71|
|2019/05/07 14:36:42 | PS Ping | 2.0.12451.0 | 23 (Threshold:20) | 2819 | 4305.09|
|2019/05/06 14:33:37 | PS Ping | 2.0.12451.0 | 24 (Threshold:20) | 2947 | 4119.65|
|2019/05/05 14:34:22 | PS Ping | 2.0.12451.0 | 22 (Threshold:20) | 2681 | 4529.3|
|2019/05/04 14:34:27 | PS Ping | 2.0.12451.0 | 23 (Threshold:20) | 2865 | 4238.09|
|2019/05/03 14:34:24 | PS Ping | 2.0.12449.0 | 22 (Threshold:20) | 2758 | 4399.92|
|2019/05/02 14:33:10 | PS Ping | 2.0.12447.0 | 23 (Threshold:20) | 2841 | 4273.55|
|2019/05/01 18:14:40 | PS Ping | 2.0.12447.0 | 22 (Threshold:20) | 2675 | 4541.99|

### Future work on throughput

@pragnagopa has been working on analyzing how to boost workers' throughput in the gRPC layer, basically around how workers should handle incoming/outgoing gRPC messages more efficiently.

This PR focused on fixing the bottleneck in the PowerShell specific logic, and the future work should shift focus to the gRPC message processing in the PowerShell worker. Open #224 to track the future work.

Also, all the measurements and analysis I did for this PR was using the default `PSWorkerInProcConcurrencyUpperBound` value, which is 1, meaning that invocations are not processed concurrently. This default behavior makes sense for the consumption plan, given the limited processing power of the VM used for consumption plan. However, for the premium plan, the VM has much better processing power, and thus the throughput should be analyzed again in that environment with the concurrency support enabled.


JustinGrote:
How would this affect exceptions that log to application insights? Seems
like it would be harder to stack trace if the function is in-memory.

TylerLeonhardt:
If we grab the latest version of PowerShell during CI, could we keep the PlatyPS check?

docs/cmdlets/Get-OutputBinding.md
```diff
@@ -13,7 +13,7 @@ Gets the hashtable of the output bindings set so far.
 ## SYNTAX
 
 ```
-Get-OutputBinding [[-Name] <String[]>] [-Purge] [<CommonParameters>]
+Get-OutputBinding [-Name <String>] [-Purge] [<CommonParameters>]
```
TylerLeonhardt: Can we go back to this being positional?

docs/cmdlets/Trace-PipelineObject.md
```diff
@@ -13,7 +13,7 @@ Writes the formatted output of the pipeline object to the information stream bef
 ## SYNTAX
 
 ```
-Trace-PipelineObject [-InputObject] <Object> [<CommonParameters>]
+Trace-PipelineObject -InputObject <Object> [<CommonParameters>]
```
TylerLeonhardt: Can me go back to this being positional?

src/FunctionInfo.cs
```diff
@@ -30,7 +31,9 @@ internal class AzFunctionInfo
         internal readonly string FuncName;
         internal readonly string EntryPoint;
         internal readonly string ScriptPath;
+        internal readonly string DeployedPSFuncName;
```
TylerLeonhardt: What's the difference between this and `FuncName` on line 31

src/Utility/Utils.cs
```diff
@@ -52,7 +59,75 @@ internal static PowerShell NewPwshInstance()
...
+                }
+            }
+
+            return pwsh;
+        }
+
+        /// <summary>
+        /// Clean up the global variables added by the function invocation.
+        /// </summary>
+        internal static void CleanupGlobalVariables(PowerShell pwsh)
```
TylerLeonhardt: This function might be useful for PSES :)

test/Unit/PowerShell/TestScripts/testBasicFunctionWithTriggerMetadata.ps1
```diff
@@ -7,5 +7,7 @@ param ($Req, $TriggerMetadata)
 
 # Used for logging tests
 Write-Verbose "a log"
+$cmdName = $MyInvocation.MyCommand.Name
```
TylerLeonhardt: Do you have a test for PSScriptRoot as well?

daxian-dbw:
> How would this affect exceptions that log to application insights? Seems
like it would be harder to stack trace if the function is in-memory.

@JustinGrote, the script block used to create the function comes from the AST parsed from the script file, such as `run.ps1`, and thus the extent that will be used for error reporting should be the same as when you invoke the file directly. So I don't think this will affect the existing error reporting. That said, we don't handle error reporting well enough today (script stack trace is not presented to the user). We have a separate issue tracking that.

docs/cmdlets/Get-OutputBinding.md
daxian-dbw: Sure, will make the change.

docs/cmdlets/Trace-PipelineObject.md
daxian-dbw: Yes, will make the change.

src/FunctionInfo.cs
daxian-dbw: > DeployedPSFuncName = $"_{FuncName}_";

This is to avoid unexpected conflicts, which should be rare, but possible to happen.

test/Unit/PowerShell/TestScripts/testBasicFunctionWithTriggerMetadata.ps1
daxian-dbw: I forgot if I have a test. Will double check and add one if there isn't.

src/FunctionLoader.cs
```diff
@@ -40,9 +40,25 @@ internal AzFunctionInfo GetFunctionInfo(string functionId)
         /// This method runs once per 'FunctionLoadRequest' during the code start of the worker.
         /// It will always run synchronously because we process 'FunctionLoadRequest' synchronously.
         /// </summary>
-        internal void LoadFunction(FunctionLoadRequest request)
+        internal static void LoadFunction(FunctionLoadRequest request)
         {
```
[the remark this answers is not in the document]
daxian-dbw: That will be a non-trivial refactoring. Can you please open an issue describing what worker is supposed to do when receiving `FileChangeEventRequest`? For example, if there are on-going invocations, should the worker stop them or wait for them to finish before taking action on the `FileChangeEventRequest`? We need more information about this scenario to decide the appropriate refactoring work.

docs/cmdlets/Get-OutputBinding.md
daxian-dbw: Fixed. It turns out parameter is made positional even without specifying `Position` for advanced function.

docs/cmdlets/Trace-PipelineObject.md
daxian-dbw: Fixed.

src/FunctionLoader.cs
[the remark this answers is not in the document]
daxian-dbw: If the requirement is simply reset the worker, then we already have `FunctionLoader.ClearLoadedFunctions` to clean up the loaded functions.

test/Unit/PowerShell/TestScripts/testBasicFunctionWithTriggerMetadata.ps1
daxian-dbw: Test case `InvokeFunctionWithSpecialVariableWorks` added.

src/Public/Commands/PushOutputBindingCommand.cs
```diff
@@ -0,0 +1,280 @@
...
+                case "graphWebhookSubscription": return DataCollectingBehavior.Collection;
+
+                // Be conservative on new output bindings
+                default: return DataCollectingBehavior.Singleton;
+            }
+        }
+
+        /// <summary>
+        /// Combine the new data with the existing data for a output binding with 'Collection' behavior.
+        /// Here is what this command do:
```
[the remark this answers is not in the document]
daxian-dbw: Fixed.

daxian-dbw:
> If we grab the latest version of PowerShell during CI, could we keep the PlatyPS check?

@TylerLeonhardt Yes we can do that. There is a gotcha though -- `Update-MarkdownHelp` doesn't recognize the `[SupportWildcard]` attribute for C# cmdlets, see my commit https://github.com/Azure/azure-functions-powershell-worker/pull/214/commits/4d44920a1d942abace0a20e2af374d79e01442ab
So we will have to fix it in `PlatyPS` before enabling it in CI.

TylerLeonhardt:
Interesting! Is there an issue on PlatyPS about this? If not, can you create a small one?

daxian-dbw:
@TylerLeonhardt It seems there is one: https://github.com/PowerShell/platyPS/issues/202
But I wonder why it works for scripts.

docs/cmdlets/Get-OutputBinding.md
```diff
@@ -49,12 +49,12 @@ The name of the output binding you want to get.
 Supports wildcards.
 
 ```yaml
-Type: String[]
+Type: String
```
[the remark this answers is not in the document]
daxian-dbw: We actually never supported `string[]` :) See the script blow:
```
# loading the psm1 version of the helper module
> Import-Module .\src\Modules\Microsoft.Azure.Functions.PowerShellWorker
> Push-OutputBinding -Name abc -Value "value1"
> Push-OutputBinding -Name zoo -Value value2
> Get-OutputBinding abc

Name                           Value
----                           -----
abc                            value1

> Get-OutputBinding zoo

Name                           Value
----                           -----
zoo                            value2

> Get-OutputBinding abc, zoo
# No output
```
The culprit is `$bindingName -like $Name` in this line: https://github.com/Azure/azure-functions-powershell-worker/blob/dev/src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1#L69

Given it never worked, I don't intend to make it work for string array until we have a request.

src/Utility/Utils.cs
```diff
@@ -52,7 +59,75 @@ internal static PowerShell NewPwshInstance()
                 }
             }
 
-            return PowerShell.Create(s_iss);
+            var pwsh = PowerShell.Create(s_iss);
+            if (s_globalVariables == null)
+            {
+                // Get the names of the built-in global variables
+                ICollection<PSVariable> globalVars = GetGlobalVariables(pwsh);
+                s_globalVariables = new HashSet<string>(globalVars.Count + 3, StringComparer.OrdinalIgnoreCase)
```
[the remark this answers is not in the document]
daxian-dbw: `PSCmdlet` doesn't show up in the global scope after evaluating a regular 'profile.ps1' that doesn't have a `param` block with `CmdletBinding()` attribute, so I don't put it here.
Actually, all other automatic variables that created along with invoking a script will be handled by the `GetType() != typeof(PSVaraible)` check below, because they will be of the type `LoaclVariable`.

src/Public/Commands/PushOutputBindingCommand.cs
```diff
@@ -0,0 +1,279 @@
...
+                        _outputBindings[Name] = newValue;
+                        return;
+
+                    default:
+                        throw new InvalidOperationException(
+                            string.Format(PowerShellWorkerStrings.UnrecognizedBehavior, _behavior.ToString()));
+                }
+            }
+
+            // Key already exists in _OutputBindings
```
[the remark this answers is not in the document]
daxian-dbw: Fixed.

test/Unit/Modules/HelperModuleTests.cs
```diff
@@ -0,0 +1,309 @@
...
+            var funcDirectory = Path.Join(AppDomain.CurrentDomain.BaseDirectory, "TestScripts", "PowerShell");
+            var rpcFuncMetadata = new RpcFunctionMetadata()
+            {
+                Name = "TestFuncApp",
+                Directory = funcDirectory,
+                ScriptFile = Path.Join(funcDirectory, "testBasicFunction.ps1"),
+                EntryPoint = string.Empty,
+                Bindings =
+                {
+                    { "req" , new BindingInfo { Direction = BindingInfo.Types.Direction.In, Type = "httpTrigger" } },
```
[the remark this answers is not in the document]
daxian-dbw: Make it line up the others.

docs/cmdlets/Get-OutputBinding.md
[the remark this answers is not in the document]
TylerLeonhardt: Sounds good!

src/Utility/Utils.cs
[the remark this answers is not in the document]
TylerLeonhardt: I see, gotcha!

test/Unit/PowerShell/PowerShellManagerTests.cs
```diff
@@ -50,132 +77,175 @@ internal static PowerShellManager NewTestPowerShellManager(ConsoleLogger logger,
...
-        internal static void Break()
+        public void Dispose()
         {
-            while (!System.Diagnostics.Debugger.IsAttached)
+            FunctionLoader.ClearLoadedFunctions();
+            s_testLogger.FullLog.Clear();
+        }
+
+        [Fact]
+        public void InvokeBasicFunctionWorks()
```
TylerLeonhardt: I feel like these Facts could be refactored into a Theory with multiple test cases since it's mostly just a chance in the ps1 that gets processed and the expected result.

I won't block on this though.

daxian-dbw:
I have updated the PR description with the detailed measurement data collected with the current changes in this PR.
/cc @pragnagopa @TylerLeonhardt @Francisco-Gamino @SatishRanjan

test/Unit/PowerShell/PowerShellManagerTests.cs
daxian-dbw: Sound good. Issue #222 was opened to track this.

src/Utility/Utils.cs
Francisco-Gamino: Could you please add a comment on what 3 means here?

docs/cmdlets/Get-OutputBinding.md
```diff
@@ -49,12 +49,12 @@ The name of the output binding you want to get.
 Supports wildcards.
 
 ```yaml
-Type: String[]
+Type: String
 Parameter Sets: (All)
 Aliases:
 
 Required: False
-Position: 1
```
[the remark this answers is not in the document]
daxian-dbw: The position is actually 0 (see the code snippet I copied from `GetOutputBindingCommand.cs` below).
Maybe it's a bug in PlatyPS.

```
        [Parameter(Position = 0, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true)]
        [SupportsWildcards]
        [ValidateNotNullOrEmpty]
        public string Name { get; set; } = "*";
```

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psd1
```diff
@@ -30,7 +30,7 @@ Copyright = '(c) Microsoft Corporation. All rights reserved.'
 Description = 'The module used in an Azure Functions environment for setting and retrieving Output Bindings.'
 
 # Minimum version of the PowerShell engine required by this module
-PowerShellVersion = '5.1'
+PowerShellVersion = '6.2'
 
 # Modules that must be imported into the global environment prior to importing this module
 RequiredModules = @()
```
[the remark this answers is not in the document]
daxian-dbw: Not sure if we should remove the unused entries, but the cleanup can be deferred to a separate PR.

src/Utility/Utils.cs
daxian-dbw: The comment is actually available right below this line :)

```
// These 3 variables are not in the built-in variables in a fresh Runspace,
// but they show up after we evaluate the 'profile.ps1' in the global scope.
```

Francisco-Gamino:
LGTM!

src/RequestProcessor.cs
```diff
@@ -38,7 +37,6 @@ internal RequestProcessor(MessagingStream msgStream)
         {
```
SatishRanjan: In the we're still awaiting for the rpc messages to be complete, and in this case the host doesn't send a new message to the client until it receives the response, so, for example if a function1 invocation request arrives while dependency download is going on then on azure portal user will see the message saying dependency download is in-progress, but if another function2 invocation request is received then host won't send the second invocation request to PowerShell language worker, and user won't see managed dependency download message on the azure portal, is it possible for PowerShell worker to receive and process invocation request concurrently?

 internal async Task ProcessRequestLoop()

        internal async Task ProcessRequestLoop()

        {

        {

            StreamingMessage request, response;

            StreamingMessage request, response;

            while (await _msgStream.MoveNext())

daxian-dbw: Yes, with the current implementation, the second invocation request will be blocked until the dependency download finishes, even if the user set `PSWorkerInProcConcurrencyUpperBound` greater than 1.
It can be addressed by not blocking the main thread but create a task to call `ProcessInvocationRequest` for each `InvocationRequest` message. The was one of my designs back when I worked on the in-proc concurrency, but that design resulted in heavy lock contention and I didn't use it in the end. See the **Design Change** section in the PR description of https://github.com/Azure/azure-functions-powershell-worker/pull/123. Quoted here:

> The original design is to not block the main thread -- it reads a request message, start a task for processing and forget about it, then it reads the next message.
With this design, every in-coming request will be handled immediately -- a task will be created for it. So, a large amount of tasks will be created rapidly, and thus a large amount of thread-pool threads will be assigned to run those tasks. All those threads will compete for an idle PowerShellManager from the pool, which results in lock contention in the pool. The average time for fetching an idle PowerShellManager instance gets worse dramatically when velocity of in-coming requests reaches about 70/s -- it reaches 994 ms in my measurements, and becomes a huge bottleneck.

I guess we can revisit this design and consider the `Rx.net`, but the competition on acquiring a `PowerShellManager` instance from the pool will continue to be the source of lock contention.

Or, think it out of the box, what if we can change the design of dependency management? Separate it from the worker and make it a service or something and can be reused by all languages (e.g. the retry logic shouldn't be implemented once per worker, right?), and make the portal more aware of the dependency management component. That would be the more long term solution to all the issues to the dependency management feature.

pragnagopa:
fyi @alrod - Once these changes are checked in, our perf test should reflect the improvements as well.

src/RequestProcessor.cs
daxian-dbw: This is definitely something worth more discussion, but I think it's out scope for this PR.
@SatishRanjan can you please create a more general issue for the challenges of the current dependency management design? @TylerLeonhardt and I talked about this a couple times, and we both think separating the dependency management logic from the worker would be the long term solution and we should start moving toward that.

SatishRanjan: @daxian-dbw - currently since there's await _msgStream.MoveNext() to move on next message from host to worker hence host does not even send invocation request to the powershell worker runtime, and I was thinking if we can remove wait on _msgStream.MoveNext() and concurrently process the invocation request, so if we do this way then the second invocation request will get dependency download in progress message as well?

daxian-dbw: @SatishRanjan I don't think we should discuss the design/fix for another issue within this PR review. As I mentioned above, I suggest you to open a new issue and discussion can happen there.

SatishRanjan: @daxian-dbw => I was expecting adding capability to process multiple invocation request concurrently was the scope of this throughput improvement change. But yes, if we want to take it separately then that should be fine as well.

SatishRanjan:
I tested managed dependency download and this feature is working as expected with this change.

src/PowerShell/PowerShellManager.cs
```diff
@@ -106,6 +123,27 @@ private void RegisterStreamEvents()
             _pwsh.Streams.Warning.DataAdding += streamHandler.WarningDataAdding;
         }
 
+        /// <summary>
+        /// Create the PowerShell function that is equivalent to the 'scriptFile' when possible.
+        /// </summary>
+        private void DeployAzFunctionToRunspace()
```
AnatoliB: PS scripts often use $MyInvocation or $PSCommandPath to locate files based on the current script file location. Will this still work?

AnatoliB: Should we give the user a way to opt out of this optimization (for this and similar reasons)? One workaround I see now is to just insert `#Requires`, perhaps we could come up with something more explicit.

src/RequestProcessor.cs
daxian-dbw: @SatishRanjan I think you misunderstood the issue your brought up here.

- First, the capability to process multiple invocations concurrently has been added to the PowerShell worker for quite a while (in Jan.), see #123. Processing multiple invocations concurrently works fine after the dependency downloading finishes. The concurrency level is controlled by the env variable `PSWorkerInProcConcurrencyUpperBound` which the user can set in the application settings.

- Second, what you brought up here is a very specific scenario -- multiple invocation triggered while the dependency downloading is in progress. By definition, no invocation can actually be served when the downloading is still in progress, so **what you really intend to address here** is to make all invocations receive the "downloading in progress" logging message.

- Third, also the most important one, the throughput should always be measured when the application is running in steady state, which means the cold start, where the downloading happens, is not in consideration when talking about throughput. Please take a look at the `Testing environment` section in my PR description, which I made it very clear that each measure is done after warming up the host/worker (`WCAT` will do another 30 seconds warming up).

src/PowerShell/PowerShellManager.cs
AnatoliB: The content of the `$MyInvocation` variable seems to be quite different before and after this change. Even though `$PSScriptRoot` and `$PSCommandPath` are identical, `$MyInvocation.MyCommand.Path` used to contain the path to the script, and now this field is absent. There are many other changes as well - I can provide more details if you need them.

This is a potentially breaking change for existing customers. If you decide to proceed with this optimization by default anyway, consider documenting this and/or providing an opt-out mechanism.

daxian-dbw: Right, documentation needs to be updated. The issue #225 was opened to track the documentation updates.

The change in `$MyInvocation.MyCommand` is by design, it's was pointing to a `ExternalScriptInfo` object previously because we were calling `run.ps1` directly, and now it's pointing to a `FunctionInfo` object, which doesn't have the `Path` property,

It's a breaking change that should definitely be called out in the documentation. But I don't think we need to provide an opt-out mechanism, partially because we are still in public preview and partially because it's something easy to change for the user.

daxian-dbw:
@SatishRanjan @AnatoliB I have addressed your comments. Please let me know if you have additional concerns. Otherwise, I will merge this PR. Thanks!

AnatoliB:
@daxian-dbw I'm ok.

daxian-dbw:
Thanks everyone for your review!

Issue #218, opened 2019-05-09
Better error handling on saving Az modules


glennamanns:
I just received a CRI (119457109) from a customer whose application went down due to the following error:
_Exception: Fail to install function app dependencies. Restarting the app may resolve the error. Error: 'The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Unable to save the module 'Az'.'_

Restarting 5 times fixed the error. Ideally, our response to them is not for them to manually restart their application. Perhaps we should implement some retry logic that better handles module download failures.

SatishRanjan:
Assigning this to @Francisco-Gamino for adding retry when downloading and saving Az modules, currently the workaround is to restart the function app.

joeyaiello:
Retry could succeed if it's a network problem, but ideally we should figure out root cause of the failure so it doesn't happen. A single download takes long enough as it is, 5 will be a little crazy.

Francisco-Gamino:
This issue has been addressed in PR #230.

eosfor:
This is still an issue looks like

![Image](https://github.com/user-attachments/assets/047aaf42-ad98-45bf-accd-757633893af3)

```pwsh
# This file enables modules to be automatically managed by the Functions service.
# See https://aka.ms/functionsmanageddependency for additional information.
#
@{
    # For latest supported version, go to 'https://www.powershellgallery.com/packages/Az'. Uncomment the next line and replace the MAJOR_VERSION, e.g., 'Az' = '5.*'
    'Az' = '12.*';
    'AzTable' = '2.*'
}
```

PR #230, opened 2019-05-22
Adding function app dependencies installation retry logic with exponential back off.

Francisco-Gamino:
This PR addresses issue https://github.com/Azure/azure-functions-powershell-worker/issues/218.

Changes in this PR:

* Adding function app dependencies installation retry logic with exponential back off. If an issue is encountered. We retry a maximum of 3 times, and we wait in between tries 1, 2, 4 seconds, respectively. 

* Before we installed the function app dependencies, we need to ensure that the user has write access to their storage account. To validate this, we create the module destination path if it does not exist, and if it exists, we try to empty it. If an issue is encountered, we error out and advice to user to ensure that they have write access to the path where the dependencies will be installed. 

* Updated error messages to keep consistency.

### DIFF
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -69,6 +69,9 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         // If we do, we use it to clean up the module destination path.
         private bool _shouldUpdateFunctionAppDependencies;
 
+        // Maximum number of tries for retry logic when installing function app dependencies.
+        private const int MaxNumberOfTries = 3;
+
         internal DependencyManager()
         {
             Dependencies = new List<DependencyInfo>();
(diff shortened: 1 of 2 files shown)


src/DependencyManagement/DependencyManager.cs
```diff
@@ -185,58 +188,86 @@ internal void Initialize(FunctionLoadRequest request)
...
+                            if (tries >= MaxNumberOfTries)
+                            {
+                                var errorMsg = string.Format(PowerShellWorkerStrings.FailToInstallFuncAppDependencies, e.Message);
+                                _dependencyError = new DependencyInstallationException(errorMsg, e);
+
+                                throw _dependencyError;
+                            }
+
+                            // Update the retry counter
+                            tries++;
```
AnatoliB: I think we should trace the errors in any case, even if we are going to retry.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -185,58 +188,86 @@ internal void Initialize(FunctionLoadRequest request)
...
+                        }
+
+                        // Wait for 2^(tries-1) seconds between retries. In this case, it would be 1, 2, and 4 seconds, respectively.
+                        double waitTimeInSeconds = Math.Pow(2, tries - 1);
+                        TimeSpan.FromSeconds(waitTimeInSeconds);
```
AnatoliB: There is no wait happening here: this line just calculates the timespan and discards it.

AnatoliB:
Consider adding unit tests.

src/DependencyManagement/DependencyManager.cs
Francisco-Gamino: I agree. Fixed.

src/DependencyManagement/DependencyManager.cs
Francisco-Gamino: Thanks for pointing this out. I lost a line of code when doing the three way merge. Fixed.

Francisco-Gamino:
> Consider adding unit tests.

We currently do not have unit test that validates the actual download logic. Perhaps we should create a small service (Azure Function) to which we redirect the calls for Save-Module as part of work for Managed Dependencies VNext.

AnatoliB:
> We currently do not have unit test that validates the actual download logic. Perhaps we should create a small service (Azure Function) to which we redirect the calls for Save-Module as part of work for Managed Dependencies VNext.

For unit testing purposes, I would not recommend interacting with a service. I would rather mock the Save-Module call out (and perhaps file system operations as well), and just make sure the mock is invoked correctly. No need to test Save-Module, but there this code contains enough complexity to deserve unit testing.

Francisco-Gamino:
> > We currently do not have unit test that validates the actual download logic. Perhaps we should create a small service (Azure Function) to which we redirect the calls for Save-Module as part of work for Managed Dependencies VNext.
> 
> For unit testing purposes, I would not recommend interacting with a service. I would rather mock the Save-Module call out (and perhaps file system operations as well), and just make sure the mock is invoked correctly. No need to test Save-Module, but there this code contains enough complexity to deserve unit testing.

@AnatoliB : As per our conversation, I've opened an issue (https://github.com/Azure/azure-functions-powershell-worker/issues/231) to add tests to mock the functionality of Save-Help as well as the file system operation.

TylerLeonhardt:
@Francisco-Gamino it's pretty easy to mock a module. You can override functions in PowerShell:

Screen Shot 2019-05-23 at 3 23 58 PM

Francisco-Gamino:
> Approved assuming that you will add unit tests for this code in a separate PR.

Thanks @AnatoliB . Will do.

Francisco-Gamino:
@anirudhgarg for FYI.

PR #237, opened 2019-06-04
Fixing faulty logic in InstallFunctionAppDependencies

Francisco-Gamino:
Fixing faulty logic in InstallFunctionAppDependencies, so that if an exception thrown, it will get caught, logged, and wrapped in a DependencyInstallationException for post processing. 

In the InstallFunctionAppDependencies, we perform two operations: 
1) Prepare the folder destination path where the function app dependencies will be installed.
* If the destination folder exist, we try to empty it
* If the folder does not exist, we create it

2) Call Save-Module to install the function app dependency

If an issue is encountered in either (1) or (2), the exception will be caught, logged and wrapped in a DependencyInstallationException for post processing. This exception is expose via DependencyManager.DependencyError

### DIFF
--- a/src/DependencyManagement/DependencyManagementUtils.cs
+++ b/src/DependencyManagement/DependencyManagementUtils.cs
@@ -53,6 +53,34 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Francisco-Gamino:
@SatishRanjan: I am going to close this issue and submit a new PR with unit tests + some changes in the DependencyManager (to accommodate the unit test).

cc: @AnatoliB

PR #263, opened 2019-06-21
Fixing test pass.

Francisco-Gamino:
Fixing test pass.


test/Unit/DependencyManagement/DependencyManagementTests.cs
```diff
@@ -452,6 +449,16 @@ protected override void RemoveSaveModuleModules(PowerShell pwsh)
         {
             return;
         }
+    }
+
+    internal class TestDependencyManager2 : DependencyManager
```
AnatoliB: Normally, I would recommend coming up with a better class name here (perhaps something referring to the idea of throwing GetLatestModuleVersion. However, I'm in the process of refactoring this code, and this name will not matter soon, so I will not insist on renaming this.

Francisco-Gamino: Sounds good. Thanks @AnatoliB.

PR #275, opened 2019-07-12
Improve module version purge on managed dependencies auto-upgrade

AnatoliB:
Improve module version purge on managed dependencies auto-upgrade

Addresses issue https://github.com/Azure/azure-functions-powershell-worker/issues/247


src/DependencyManagement/DependencyManagerStorage.cs
```diff
@@ -90,5 +90,32 @@ public DateTime GetSnapshotCreationTimeUtc(string path)
...
+        public void SetSnapshotAccessTimeToUtcNow(string path)
+        {
+            var markerFilePath = DependencySnapshotFolderNameTools.CreateLastAccessMarkerFilePath(path);
+            if (File.Exists(markerFilePath))
+            {
+                File.SetLastWriteTimeUtc(markerFilePath, DateTime.UtcNow);
+            }
+            else
+            {
+                var file = File.Open(markerFilePath, FileMode.Create, FileAccess.Write, FileShare.ReadWrite | FileShare.Delete);
```
Francisco-Gamino: Should we use a ```using statement``` to ensure that the file is close in case of an exception occurs while doing File.Open(..)?

docs/designs/PowerShell-AzF-Overall-Design.md
```diff
@@ -494,7 +494,9 @@ However, if the latest snapshot is _not acceptable_ (i.e. it does not contain mo
...
 
-Incomplete and old snapshots that are no longer in use are periodically removed from the file storage. No other changes are ever performed to the snapshots that were once installed.
+Incomplete and old snapshots that are no longer in use are periodically removed from the file storage. In order to allow detecting unused snapshots, each PowerShell worker keeps "touching" a file named `.used` at the root of the used snapshot folder every `PSWorkerHeartbeatPeriodMinutes` minutes. Before and after installing any new snapshot, every PowerShell worker looks for unused snapshots by checking the folder creation time and the `.used` file modification time. If both these time values are older than (`PSWorkerHeartbeatPeriodMinutes` + `PSWorkerOldSnapshotHeartbeatMarginMinutes`) minutes, the snapshot is considered unused, so the PowerShell worker removes it. The latest `PSWorkerMinNumberOfSnapshotsToKeep` snapshots will never be removed, regardless of usage. (`PSWorkerHeartbeatPeriodMinutes`, `PSWorkerOldSnapshotHeartbeatMarginMinutes`, and `PSWorkerMinNumberOfSnapshotsToKeep` are environment variables configurable via Application Settings of a Function App.)
```
Francisco-Gamino: We should close with @eamonoreilly on the default values for ```PSWorkerMinNumberOfSnapshotsToKeep```, ```PSWorkerHeartbeatPeriodMinutes``` and ```PSWorkerOldSnapshotHeartbeatMarginMinutes```.

src/DependencyManagement/DependencyManagerStorage.cs
AnatoliB: If `File.Open` throws, it does not return anything, so `using` will not be able to call `.Dispose` anyway, and the behavior will be exactly the same. This is just a matter of taste: I found that `using` with an empty body invites more questions than just explicitly invoking `Dispose`.

docs/designs/PowerShell-AzF-Overall-Design.md
AnatoliB: Yes, and on the names too, as they will be exposed to the user.

src/DependencyManagement/DependencyManagerStorage.cs
Francisco-Gamino: Sounds good. We'll keep it as is.

docs/designs/PowerShell-AzF-Overall-Design.md
Francisco-Gamino: Since this feature has not been released, we can merge it as it is right now. Then, once we close on the name of the settings, we can update the code.

PR #290, opened 2019-07-24
Remove partially installed snapshots

AnatoliB:
If installing a managed dependencies snapshot fails for any reason, the current code just abandons the snapshot folder, relying on it to be eventually removed by regular purge. As a result, the partially installed snapshot keeps occupying file space, and prevents the next installation attempt from starting soon. There is no easy way to avoid these consequences when the PowerShell worker process is terminated. However, if the worker keeps running and it is aware of the installation attempt failure, it has an opportunity to perform the clean up sooner. This is what this PR achieves.

### DIFF
--- a/test/Unit/DependencyManagement/DependencyManagementTests.cs
+++ b/test/Unit/DependencyManagement/DependencyManagementTests.cs
@@ -336,24 +336,27 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Test
 
                     var relevantLogs = _testLogger.FullLog.Where(
                         message => message.StartsWith("Error: Fail to install module")
-                                   || message.StartsWith("Trace: Installing FunctionApp dependent modules")).ToList();
+                                   || message.StartsWith("Trace: Installing FunctionApp dependent modules")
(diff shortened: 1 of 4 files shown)


test/Unit/DependencyManagement/DependencyManagementTests.cs
```diff
@@ -336,24 +336,27 @@ public void TestManagedDependencyRetryLogicMaxNumberOfTries()
 
                     var relevantLogs = _testLogger.FullLog.Where(
                         message => message.StartsWith("Error: Fail to install module")
-                                   || message.StartsWith("Trace: Installing FunctionApp dependent modules")).ToList();
+                                   || message.StartsWith("Trace: Installing FunctionApp dependent modules")
+                                   || message.StartsWith("Warning: Failed to install dependencies")).ToList();
 
                     // Here we will get four logs: one that says that we are installing the
```
Francisco-Gamino: Please update the comment to reflect the code.

AnatoliB: Thank you, updated.

PR #293, opened 2019-07-24
Shorten managed dependencies snapshot folder names

AnatoliB:
Shorten managed dependencies snapshot folder names:
- Remove dots and dashes.
- Remove microseconds.
- Use just two last digits for the year.


src/DependencyManagement/DependencySnapshotFolderNameTools.cs
```diff
@@ -29,11 +29,11 @@ public static string CreateUniqueName()
         /// appending a postfix, so that that the resulting path follows a different
         /// pattern and can be discovered using a different file mask.
         /// For example, for the _installed_ path
-        ///     ".../20190710-1234.567890.r"
+        ///     ".../1907101234567r"
         /// the _installing_ path will be:
-        ///     ".../20190710-1234.567890.r.i"
-        /// This makes it possible to enumerate all the installed snapshots by using ".../*.r" mask,
-        /// and all the installing snapshots by using ".../*.i" mask.
+        ///     ".../1907101234567ri"
```
Francisco-Gamino: Should this be ```1907101234567i```?

Francisco-Gamino: @AnatoliB: Please rebase your code. The unsuccessful checks should pass now :).

AnatoliB: @Francisco-Gamino 
> Should this be 1907101234567i?
It could, but the code will actually do `1907101234567ri`, for the sake of simplifying the transformation from the "installed" name to the "installing" name. Currently, we just append `i`, which is a no-brainer. If we wanted to do `1907101234567i` instead, we would have to do a replacement, make a decision on how to handle the situation when the original name does not end with "i". Though not very difficult, these decisions complicate the problem without a valuable reason.

Francisco-Gamino: That makes sense. Thanks for the clarification @AnatoliB.

PR #295, opened 2019-07-24
Changed environment variables related to managed dependencies

AnatoliB:
Changed environment variables related to managed dependencies:
- The `PSWorker` prefix is replaced with `MD`.
- The variables containing time intervals are now expected to contain strings that can be parsed by `System.TimeSpan.Parse` (for example, `00:30:00` means "30 minutes"); the word `Minutes` is also removed from the names.


Francisco-Gamino:
Please rebase your code, after that, the checks should pass. Thanks!

PR #314, opened 2019-08-23
Refactoring: Retrieve all PSWorker config values via PowerShellWorkerConfiguration

AnatoliB:
Refactoring: Retrieve all PSWorker config values via PowerShellWorkerConfiguration

### DIFF
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -19,7 +19,8 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
     {
         #region Private fields
 
-        private static readonly TimeSpan s_minBackgroundUpgradePeriod = GetMinBackgroundUpgradePeriod();
+        private static readonly TimeSpan s_minBackgroundUpgradePeriod =
+            PowerShellWorkerConfiguration.GetTimeSpan("MDMinBackgroundUpgradePeriod") ?? TimeSpan.FromDays(1);
 
         private readonly IDependencyManagerStorage _storage;
 
@@ -263,12 +264,6 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
             }
 
             return Path.GetFullPath(Path.Join(requestMetadataDirectory, ".."));
-        }
-
-        private static TimeSpan GetMinBackgroundUpgradePeriod()
-        {
-            var value = Environment.GetEnvironmentVariable("MDMinBackgroundUpgradePeriod");
-            return string.IsNullOrEmpty(value) ? TimeSpan.FromDays(1) : TimeSpan.Parse(value);
         }
 
         #endregion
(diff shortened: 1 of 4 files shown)


src/Utility/PowerShellWorkerConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Utility
+{
+    using System;
+
+    internal static class PowerShellWorkerConfiguration
+    {
+        public static string GetString(string name)
```
Francisco-Gamino: Since these are app settings, would it make sense to add validation to the ```string name```? This way we can detect invalid customer inputs.

src/DependencyManagement/DependencySnapshotPurger.cs
```diff
@@ -128,34 +128,17 @@ private DateTime GetSnapshotAccessTimeUtc(string path, ILogger logger)
 
         private static TimeSpan GetHeartbeatPeriod()
         {
-            return GetEnvironmentVariableTimeSpanValue("MDHeartbeatPeriod") ?? TimeSpan.FromMinutes(60);
+            return PowerShellWorkerConfiguration.GetTimeSpan("MDHeartbeatPeriod") ?? TimeSpan.FromMinutes(60);
```
Francisco-Gamino: Should we move all default values to constants?

src/DependencyManagement/DependencyManager.cs
```diff
@@ -19,7 +19,8 @@ internal class DependencyManager : IDisposable
     {
         #region Private fields
 
-        private static readonly TimeSpan s_minBackgroundUpgradePeriod = GetMinBackgroundUpgradePeriod();
+        private static readonly TimeSpan s_minBackgroundUpgradePeriod =
+            PowerShellWorkerConfiguration.GetTimeSpan("MDMinBackgroundUpgradePeriod") ?? TimeSpan.FromDays(1);
```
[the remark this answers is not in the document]
AnatoliB: Yes, we'll update the docs soon.

src/Utility/PowerShellWorkerConfiguration.cs
AnatoliB: The `name` value here always comes from our code, not from the customer. The `PowerShellWorkerConfiguration` class implementation always retrieves them from environment variables, so it is not even aware of the concept of app settings.

We may want to perform this validation in the code that copies app settings to environment variables, though. However, this validation is not free, so let's do it when and if we see the need.

src/DependencyManagement/DependencySnapshotPurger.cs
AnatoliB: What I like about the current code is that the names, the types, and the default values are all on the same line, which makes it super-obvious and easy to maintain. I've seen code having constants for default values in similar cases, and I believe this is an anti-pattern: it introduces a level of indirection that makes values harder to look up, and creates an opportunity for copy-paste bugs, without providing valuable benefits.

src/Utility/PowerShellWorkerConfiguration.cs
Francisco-Gamino: Sounds good. Thanks for clarifying.

src/DependencyManagement/DependencySnapshotPurger.cs
Francisco-Gamino: I do see your point. I am fine if you'd like to keep it as is.

PR #319, opened 2019-09-06
Log a warning when a function invocation is blocked because of the concurrency limit

AnatoliB:


src/resources/PowerShellWorkerStrings.resx
```diff
@@ -271,4 +271,7 @@
   <data name="PromotedDependencySnapshot" xml:space="preserve">
     <value>Promoted dependency snapshot '{0}' to '{1}'.</value>
   </data>
+  <data name="FunctionQueuingRequest" xml:space="preserve">
+    <value>Your function '{0}' is queuing requests as there are no available runspaces. You may be able to increase your throughput by following the best practices on https://aka.ms/functions-powershell-concurrency.</value>
```
Francisco-Gamino: Should add this uri ```https://aka.ms/functions-powershell-concurrency``` pragmatically when the  ```FunctionQueuingRequest``` is created?

AnatoliB: We can, but why? Would you just have a constant in the code (how is this better?), or you are talking about taking it from somewhere else?

Issue #292, opened 2019-07-24
Support "RpcHttpBodyOnly"capability


yojagad:
Host has a "RpcHttpBodyOnly" capability where it does not send duplicated data through gRPC. It only sends the "body" property which is either bytes or string.

Reference PR: https://github.com/Azure/azure-functions-host/pull/4660

Powershell needs to make modifications on the worker side accordingly in order to decode it whenever necessary.

PR #323, opened 2019-09-10
Implemented "RpcHttpBodyOnly" capability

AnatoliB:
Resolving issue https://github.com/Azure/azure-functions-powershell-worker/issues/292.

The goal is to reduce data duplication in function invocation requests, which will reduce the I/O volume between the host and the PS worker, and potentially allow larger request payloads.

### DIFF
--- a/src/Utility/TypeExtensions.cs
+++ b/src/Utility/TypeExtensions.cs
@@ -54,15 +54,10 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Utility
             if (rpcHttp.Body != null)
             {
                 httpRequestContext.Body = rpcHttp.Body.ToObject();
-            }
 
-            if (rpcHttp.RawBody != null)
-            {
-                object rawBody = rpcHttp.RawBody.DataCase == TypedData.DataOneofCase.String
-                    ? rpcHttp.RawBody.String
-                    : rpcHttp.RawBody.ToObject();
-
-                httpRequestContext.RawBody = rawBody;
+                httpRequestContext.RawBody = rpcHttp.Body.DataCase == TypedData.DataOneofCase.String
+                                                ? rpcHttp.Body.String
+                                                : rpcHttp.Body.ToObject();
             }
 
             return httpRequestContext;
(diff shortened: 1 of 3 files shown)


src/Utility/TypeExtensions.cs
```diff
@@ -99,6 +90,21 @@ internal static object ToObject(this TypedData data)
...
+            switch (rpcHttpBody.DataCase)
+            {
+                case TypedData.DataOneofCase.String:
+                    return rpcHttpBody.String;
+
+                case TypedData.DataOneofCase.Bytes:
+                    return Encoding.UTF8.GetString(rpcHttpBody.Bytes.ToByteArray());
+
+                default:
+                    return rpcHttpBody.ToObject();
```
Francisco-Gamino: For the default case in which ```rpcHttpBody``` is neither a string nor bytes, then both ```httpRequestContext.Body``` and ```httpRequestContext.RawBody```  will be rpcHttp.Body.ToObject(). Is this intentional?

AnatoliB: As far as I understand, the host will be sending us either `TypedData.DataOneofCase.String` or `TypedData.DataOneofCase.Bytes` now, so the default branch will not be executed. However, we need to come up with a safe default behavior if the host sends anything else in the future. Invoking `ToObject()` seems to be a reasonable default to me, but I'm open to other suggestions.

Also, don't pay too much attention to the resulting `httpRequestContext.Body` now. This PR focuses on `RawBody` only, and the fix for https://github.com/Azure/azure-functions-powershell-worker/issues/241 is also coming soon, in a separate PR.

Francisco-Gamino: I thought about it a bit more. I was thinking that we should keep the same behavior as other workers. However, this has not been implemented in other workers.

Would it be possible to add a TODO there to check how other workers will handle this? Other than that, LGTM

AnatoliB: This is in fact implemented in the NodeJS worker (see [fromRpcHttpBody](https://github.com/Azure/azure-functions-nodejs-worker/blob/d6a0c97ef7b0142a5c737c5b80d939da060bb189/src/converters/RpcHttpConverters.ts#L40)), and it seems to be doing the same.

Francisco-Gamino: Great. Thanks for checking.

Issue #321, opened 2019-09-09
Log PowerShell Worker version on Startup


FinVamp1:
Similar to the Python Language Worker issue https://github.com/Azure/azure-functions-python-worker/issues/512 it would be good if we log a version number on Startup to help diagnose issues.

PR #330, opened 2019-09-16
Log PowerShell Worker version on start

AnatoliB:
Resolving https://github.com/Azure/azure-functions-powershell-worker/issues/321


src/Worker.cs
```diff
@@ -24,6 +26,10 @@ public static class Worker
         /// </summary>
         public async static Task Main(string[] args)
         {
+            RpcLogger.WriteSystemLog(
+                LogLevel.Information,
+                $"Azure Functions PowerShell Worker version [{typeof(Worker).Assembly.GetName().Version}]");
```
Francisco-Gamino: Please make this a string resource. Other than that, LGTM

PR #334, opened 2019-09-23
Fix log messages

AnatoliB:
Fixing typos, replacing"FunctionApp" with "function app", etc.

### DIFF
--- a/src/resources/PowerShellWorkerStrings.resx
+++ b/src/resources/PowerShellWorkerStrings.resx
@@ -202,9 +202,6 @@
   <data name="InstallingFunctionAppRequiredModules" xml:space="preserve">
     <value>Installing function app required modules to '{0}'.</value>
   </data>
-  <data name="FailToClenupModuleDestinationPath" xml:space="preserve">
-    <value>Failed to clean up module destination path '{0}'</value>
-  </data>
   <data name="LogConcurrencyUpperBound" xml:space="preserve">
     <value>The enforced concurrency level (pool size limit) is '{0}'.</value>
   </data>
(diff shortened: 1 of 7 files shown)


src/resources/PowerShellWorkerStrings.resx
```diff
@@ -194,16 +194,16 @@
...
   </data>
   <data name="FailToResolveHomeDirectory" xml:space="preserve">
-    <value>Fail to resolve '{0}' path in App Service.</value>
+    <value>Failed to resolve '{0}' path in App Service.</value>
   </data>
   <data name="InstallingFunctionAppDependentModules" xml:space="preserve">
-    <value>Installing FunctionApp dependent modules to '{0}'.</value>
+    <value>Installing function app dependent modules to '{0}'.</value>
   </data>
   <data name="FailToClenupModuleDestinationPath" xml:space="preserve">
```
Francisco-Gamino: I don't think this resource string is being used anymore. Please remove it.

AnatoliB: I can actually see it being used: https://github.com/Azure/azure-functions-powershell-worker/blob/0be9d316c580c8a9d95345b5d6432fb5478d7835/src/DependencyManagement/DependencySnapshotInstaller.cs#L47

AnatoliB: The message needs to be fixed though: these are not _dependent_ modules, it's the other way around.

Francisco-Gamino: > The message needs to be fixed though: these are not dependent modules, it's the other way around.

You are right. We should fix the message.

The string I was referring to remove is ```FailToClenupModuleDestinationPath```.

AnatoliB: Yes. Removing.

Francisco-Gamino: Thank you.

PR #335, opened 2019-09-23
Log function load and invocation performance data

AnatoliB:


src/RequestProcessor.cs
```diff
@@ -282,7 +298,7 @@ private void ProcessInvocationRequestImpl(StreamingMessage request, AzFunctionIn
                 // Invoke the function and return a hashtable of out binding data
                 Hashtable results = functionInfo.Type == AzFunctionType.OrchestrationFunction
                     ? InvokeOrchestrationFunction(psManager, functionInfo, invocationRequest)
```
Francisco-Gamino: Even though we do not support ```Durable Functions``` yet, we should add the parameter ```FunctionInvocationPerformanceStopwatch stopwatch``` to ```InvokeOrchestrationFunction``` for completeness.

src/Utility/FunctionInvocationPerformanceStopwatch.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+
+    internal class FunctionInvocationPerformanceStopwatch
+    {
+        public enum Checkpoint
+        {
+            DependenciesAvailable,
```
Francisco-Gamino: Would it make sense to add a checkpoint for when the function app dependencies are being installed? This would be for the upgrade scenario.

AnatoliB: The main purpose of this PR is to measure cold start impact. By design, _upgrading_ dependencies is _not_ on the function invocation path, as it is performed on the background, so we would not encounter this checkpoint anywhere on this invocation chain.

If we are interested in measuring how long background upgrades take, we can already do it based on existing traces ("Installing function app required modules" etc.). For the background upgrade, precision is not that important, so we can rely on the regular log timestamps. I'm doing something special here because we want _at least_ 50-100 ms precision.

src/RequestProcessor.cs
AnatoliB: Ok.

src/Utility/FunctionInvocationPerformanceStopwatch.cs
Francisco-Gamino: Good point. Thanks for clarifying.

PR #340, opened 2019-09-26
Change PS Worker version to 1.0, remove "preview"

AnatoliB:


anirudhgarg:
When do we plan to do this ?

AnatoliB:
@anirudhgarg If we want to have version 1.* released/deployed when we declare GA, we need to commit this by Wednesday. I'm planning to do this as soon as we confirm that the change is technically correct. Please let me know if you have anything else in mind.

anirudhgarg:
Thanks @AnatoliB. Sounds good.

PR #344, opened 2019-09-27
Log all the modules and versions in the promoted snapshot

AnatoliB:


src/resources/PowerShellWorkerStrings.resx
```diff
@@ -292,4 +292,10 @@
   <data name="FunctionInvocationStarting" xml:space="preserve">
     <value>Invoking function '{0}' code {1} ms after receiving request. Invocation performance details: {2}</value>
   </data>
+  <data name="DependencyShapshotContent" xml:space="preserve">
+    <value>Dependency snapshot '{0}' contains module '{1}' version '{2}'.</value>
+  </data>
+  <data name="FailedToEnumerateDependencySnapshotContent" xml:space="preserve">
+    <value>Could not enumerated dependency snapshot '{0}' content: {1}</value>
```
Francisco-Gamino: To keep the same language as the other string resources, should this be ```Failed to enumerate...```?

AnatoliB: I don't even understand how I came up with this message to begin with... :-)
Thank you for noticing, fixed.

PR #351, opened 2019-10-05
Don't remove the initial snapshot

### LINKED ISSUE #350: Missing dependency modules after initial deployment
**Impact**
If the first app deployment is followed by a rapid sequence of function invocations, some function invocations may not have the required modules available until all the workers restart. Background auto-upgrades after that are not impacted.

**Root cause**
In this situation, multiple PowerShell workers may start around the same time. By design, since there is no installed acceptable snapshot yet, each worker will start installing dependency snapshots independently. The first worker that finishes downloading its snapshot promotes it to an installed snapshot and becomes ready to handle function invocations. However, the next worker compares the snapshot it just downloaded with the latest installed snapshot and, if they are equivalent, removes its own snapshot.

The logic of removing equivalent snapshots is intentional, but it was meant to apply to _background upgrades_ only. In the background upgrade case, the worker already has an acceptable snapshot assigned, so the newly downloaded snapshot can be safely removed. However, in the initial installation case, the worker cannot be reassigned to a different snapshot, so removing the downloaded snapshot means that the worker will not have _any_ snapshot until restarted.

**Workaround**
Restart the app.

AnatoliB:
Resolves #350.

### DIFF
--- a/src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
+++ b/src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
@@ -73,7 +73,12 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
 
                 using (var pwsh = pwshFactory())
                 {
-                    _installer.InstallSnapshot(_dependencyManifest, nextSnapshotPath, pwsh, logger);
+                    _installer.InstallSnapshot(
+                        _dependencyManifest,
+                        nextSnapshotPath,
+                        pwsh,
+                        removeIfEquivalentToLatest: true,
+                        logger);
                 }
 
                 // Now that a new snapshot has been installed, there is a chance an old snapshot can be purged.
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -206,7 +206,12 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         {
             try
             {
-                _installer.InstallSnapshot(_dependenciesFromManifest, _currentSnapshotPath, pwsh, logger);
+                _installer.InstallSnapshot(
+                    _dependenciesFromManifest,
+                    _currentSnapshotPath,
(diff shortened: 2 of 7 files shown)


src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
```diff
@@ -73,7 +73,12 @@ public string InstallAndPurgeSnapshots(Func<PowerShell> pwshFactory, ILogger log
 
                 using (var pwsh = pwshFactory())
                 {
-                    _installer.InstallSnapshot(_dependencyManifest, nextSnapshotPath, pwsh, logger);
+                    _installer.InstallSnapshot(
+                        _dependencyManifest,
+                        nextSnapshotPath,
+                        pwsh,
+                        removeIfEquivalentToLatest: true,
```
Francisco-Gamino: Could you please add a note why we set this to true? ```removeIfEquivalentToLatest: true```

src/DependencyManagement/DependencyManager.cs
```diff
@@ -206,7 +206,12 @@ private void InstallSnapshotInForeground(PowerShell pwsh, ILogger logger)
         {
             try
             {
-                _installer.InstallSnapshot(_dependenciesFromManifest, _currentSnapshotPath, pwsh, logger);
+                _installer.InstallSnapshot(
+                    _dependenciesFromManifest,
+                    _currentSnapshotPath,
+                    pwsh,
+                    removeIfEquivalentToLatest: false,
```
Francisco-Gamino: Could you please add a note why we set ```removeIfEquivalentToLatest: false```?

src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
AnatoliB: Thank you, added.

src/DependencyManagement/DependencyManager.cs
AnatoliB: Done.

PR #366, opened 2019-12-02
[Durable Functions] Clean up Durable programming model

AnatoliB:
Clean up Durable programming model:

- Expect `orchestrationClient` to be an input binding instead of an output binding, for consistency with other languages.
- Rename `Invoke-ActivityFunctionAsync` to `Invoke-ActivityFunction`.
- Don't require orchestration client data to be explicitly passed as a parameter.
- Tolerate different casing of `orchestrationClient` in function.json.

### DIFF
--- a/docs/designs/PowerShell-AzF-Overall-Design.md
+++ b/docs/designs/PowerShell-AzF-Overall-Design.md
@@ -549,17 +549,17 @@ We had a prototype of Durable Functions in PowerShell worker to enable the '_Fun
 2. Make the orchestrator function able to be stopped at certain point safely.
 3. Make the PowerShell worker able to replay an orchestrator function, namely skipping the actions that are done in previous runs/replays based on the save logs.
 
-They were solved by introducing the new cmdlet "_Invoke-ActivityFunctionAsync_" plus having the worker invoke the orchestrator function with the async PowerShell API.
+They were solved by introducing the new cmdlet "_Invoke-ActivityFunction_" plus having the worker invoke the orchestrator function with the async PowerShell API.
 
 Internally, the worker shares context information with the cmdlet, including _the existing saved logs_ sent from the host about the running orchestrator function and _a wait handler_ - let's call it A - which the cmdlet will set after it triggers an activity function. The async API used to start the orchestrator function returns _another wait handler_ - let's call it B - which will be set when the invocation finishes. Then the invoking thread will call '_WaitHandler.WaitAny_' on those two wait handlers.
 
 - If the call to '_WaitAny_' returns because the wait handler A was set, then that means an activity function was just triggered, and the orchestrator function should be stopped now (it will be triggered again later after the activity function finishes). So, in this case, the invoking thread will stop the orchestrator function that is running asynchronously.
 - If the call to '_WaitAny_' returns because the wait handler B was set, then that means the orchestrator function has run to its completion.
 
-The cmdlet '_Invoke-ActivityFunctionAsync_' has the following syntax, the '_-FunctionName_' being the name of the activity function to invoke and '_-Input_' being the argument to the activity function.
+The cmdlet '_Invoke-ActivityFunction_' has the following syntax, the '_-FunctionName_' being the name of the activity function to invoke and '_-Input_' being the argument to the activity function.
 
 ```powershell
-Invoke-ActivityFunctionAsync [-FunctionName] <string> [[-Input] <Object>] [<CommonParameters>]
+Invoke-ActivityFunction [-FunctionName] <string> [[-Input] <Object>] [<CommonParameters>]
 ```
 
 When it's invoked to trigger an activity function, it first checks the existing logs shared by the worker to see if this invocation of the activity function has already done previously. If so, the cmdlet simply returns the result. If not, the cmdlet will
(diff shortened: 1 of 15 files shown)


docs/designs/PowerShell-AzF-Overall-Design.md
```diff
@@ -549,17 +549,17 @@ We had a prototype of Durable Functions in PowerShell worker to enable the '_Fun
...
 
 - If the call to '_WaitAny_' returns because the wait handler A was set, then that means an activity function was just triggered, and the orchestrator function should be stopped now (it will be triggered again later after the activity function finishes). So, in this case, the invoking thread will stop the orchestrator function that is running asynchronously.
 - If the call to '_WaitAny_' returns because the wait handler B was set, then that means the orchestrator function has run to its completion.
 
-The cmdlet '_Invoke-ActivityFunctionAsync_' has the following syntax, the '_-FunctionName_' being the name of the activity function to invoke and '_-Input_' being the argument to the activity function.
+The cmdlet '_Invoke-ActivityFunction_' has the following syntax, the '_-FunctionName_' being the name of the activity function to invoke and '_-Input_' being the argument to the activity function.
```
Francisco-Gamino, line 559: Should we add the ```FunctionName``` parameter is mandatory?
It is obvious from the ```Get-Command -Syntax``` output, but only if you are familiar with PowerShell.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
```diff
@@ -9,20 +9,25 @@ function CheckIfDurableFunctionsEnabled {
 	}
 }
 
+function GetOrchestrationClientFromModulePrivateData {
+    $PrivateData = $PSCmdlet.MyInvocation.MyCommand.Module.PrivateData
+    $PrivateData['OrchestrationClient']
```
Francisco-Gamino: We should check that ```$PrivateData``` is not null or empty.
```
if ($PrivateData)
{
   $PrivateData['OrchestrationClient']
}
```

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
```diff
@@ -37,34 +42,31 @@ function Start-NewOrchestration {
         [Parameter(
             Position=2,
```
Francisco-Gamino: IMPO, I believe that using a parameter by name makes more sense that having a positional parameter. This way is easier to understand what's the intention of the user.

src/Durable/SetFunctionInvocationContext.cs
```diff
@@ -17,16 +17,16 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
...
         /// <summary>
-        /// The orchestration client output binding name.
+        /// The orchestration client.
         /// </summary>
-        [Parameter(Mandatory = true, ParameterSetName = StarterKey)]
-        public string OrchestrationStarter { get; set; }
+        [Parameter(Mandatory = true, ParameterSetName = OrchestrationClientKey)]
+        public object OrchestrationClient { get; set; }
 
         [Parameter(Mandatory = true, ParameterSetName = "Clear")]
```
Francisco-Gamino, line 31: Sorry I missed this on my previous review. Why is this switch mandatory?

docs/designs/PowerShell-AzF-Overall-Design.md
AnatoliB: This will be the right thing to do in a reference or a tutorial, so let's do it there. But we are in a design doc now, so I would keep this detail out.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
AnatoliB: Thank you, done.

docs/designs/PowerShell-AzF-Overall-Design.md
Francisco-Gamino: Sounds good. Thanks.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
AnatoliB: I think it's a good practice to _enable_ positional parameters for one or two most obvious things. For example, you wouldn't expect `Invoke-RestMethod` to require specifying the Uri parameter name - what else could it possibly be? And, if the name were mandatory, some users would struggle to remember: was it "Uri", "Url", or something else? I feel that `FunctionName` and `InputObject` here are in the same category: the user would intuitively expect these parameters in this order, without having to remember whether this is "FunctionName", or just "Function", or just "Name". This is also important because we may not have intellisense or autocomplete for these commands. And then it's up to the user whether to specify parameter names or not.

One thing I want to fix though is change Position=2 to 1. No idea why it was 2 - leftovers of legacy code.

src/Durable/SetFunctionInvocationContext.cs
AnatoliB: No idea, fixing.

AnatoliB: Oh, no, I understand now. This is because this switch is used for selecting the parameter set: for this parameter set ("Clear"), this switch is always mandatory because this is what makes PowerShell select this parameter set. If this switch is not provided, a different parameter set will be used. No fix required. But good catch. :-)

Francisco-Gamino: Got it. Thanks for clarifying.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
Francisco-Gamino: That makes sense. Thanks for the clarification. Let's keep it as is.

Francisco-Gamino (approved):
LGTM

PR #378, opened 2020-01-16
Change PS Worker version to 3

AnatoliB:
We are planning to release PowerShell Worker for Functions v3 from branches based on the **dev** branch, so setting the worker version to 3 to match Functions version numbers.


Francisco-Gamino (approved):
LGTM

PR #381, opened 2020-01-17
v3/ps7 branch setup

AnatoliB:
Setting up a PowerShell Worker build that will be used for Functions v3 and PowerShell 7.
This PR just sets up a separate build, not actually migrating to PS7 yet.


Francisco-Gamino (approved):
LGTM

PR #382, opened 2020-01-21
Return dev branch build

AnatoliB:
We want to be able to keep `dev` and `v3.x/ps7` branches in sync with each other by using simple fast-forward merges (in the same way as we did with `dev` and `master` before), so we keep the history clean. In order to do that, we'll need to have both `dev` and `v3.x/ps7` builds here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,12 @@
 
 |Branch|Status|
 |---|---|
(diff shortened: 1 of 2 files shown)


Francisco-Gamino (approved):
LGTM

PR #383, opened 2020-01-22
Migrate PowerShell worker to PowerShell 7

AnatoliB:
Migrate PowerShell worker to PowerShell 7:
- Using PowerShell SDK **v7.0.0-rc.2** for now. We'll move to the GA release as soon as it is available.
- In PowerShell 7, the default error formatting has been intentionally changed, and a new `Get-Error` cmdlet has been introduced. The new cmdlet seems to provide all the details that our custom error formatting code provides (such as call stacks, inner exceptions, etc.), so we should consider aligning with PS7 and removing our custom code, but we'll do it separately. For now, this PR just switches to the PS6-style error formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ On macOS if you installed via `brew`
 
 Copy the result of the `publish` directory into a `powershell` folder under `workers`:
 ```powershell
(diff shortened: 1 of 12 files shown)


Francisco-Gamino (approved):
LGTM

PR #387, opened 2020-01-26
Make E2E tests run on Functions V3 runtime

AnatoliB:
We'll need to propagate this change to both `v3.x/ps7` and `v3.x/ps6` branches.

### DIFF
--- a/test/E2E/TestFunctionApp/extensions.csproj
+++ b/test/E2E/TestFunctionApp/extensions.csproj
@@ -1,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+	<WarningsAsErrors></WarningsAsErrors>
+	<DefaultItemExcludes>**</DefaultItemExcludes>
+  </PropertyGroup>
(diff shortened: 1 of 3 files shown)


test/E2E/TestFunctionApp/extensions.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+	<WarningsAsErrors></WarningsAsErrors>
```
Francisco-Gamino, line 4: If we are not enabling `WarningsAsErrors`, should we remove it?

Francisco-Gamino (approved):
One minor comment. Other than that, LGTM.

test/E2E/TestFunctionApp/extensions.csproj
AnatoliB: This code was auto-generated by the `func extensions install` command, so I would suggest we leave it as is, even though we could technically improve it.

Francisco-Gamino: Sounds good. Thanks for clarifying.

PR #388, opened 2020-01-26
Move PS7 worker to a version-specific subfolder

AnatoliB:
We'll need a similar change on the `v3.x/ps6` branch.


Francisco-Gamino (approved):
LGTM

PR #394, opened 2020-01-29
Fix sanitizeRuntimeVersionRegex entry in worker config

AnatoliB:
The `sanitizeRuntimeVersion` property has been renamed to `sanitizeRuntimeVersionRegex` in the Functions Host code, so fixing it in the worker config as well.


Francisco-Gamino (approved):
LGTM

PR #395, opened 2020-01-30
Use Get-Error instead of custom error formatting logic

### LINKED ISSUE #384: For PowerShell 7, use the new Get-Error cmdlet instead of the custom error formatting logic
Compare the content of error reports produced by the custom logic implemented in the PowerShell worker with the content the new Get-Error cmdlet produces. If all the relevant data is present, just replace the custom code with a Get-Error invocation. If something important is missing, consider filing a PowerShell issue instead.

AnatoliB:
Resolves #384

### DIFF
--- a/src/PowerShell/ErrorRecordFormatter.cs
+++ b/src/PowerShell/ErrorRecordFormatter.cs
@@ -13,7 +13,6 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.PowerShell
     internal class ErrorRecordFormatter
     {
         private const string TruncationPostfix = "...";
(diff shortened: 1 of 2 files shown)


Francisco-Gamino (approved):
LGTM

Francisco-Gamino:
Code change looks good.
Quick question: This change only applies to PS7, right?

AnatoliB:
> Code change looks good.
> Quick question: This change only applies to PS7, right?

Correct. So, this change is going to the `dev` branch, and subsequently to `v3.x/ps7`, but will never go to `v2.x` nor ` v3.x/ps6`.

PR #396, opened 2020-01-31
Log PowerShell version

AnatoliB:
We'll need this in both `v3.x/ps6` and `v3.x/ps7` branches. It is not strictly necessary in v2, so I will _not_ propagate it to the `v2.x` branch at this point.


Francisco-Gamino (approved):
LGTM.

Francisco-Gamino:
Just curious if you happen to have an example of how the version information would look like in the logs?

AnatoliB:
> Just curious if you happen to have an example of how the version information would look like in the logs?

It will be the same as what you would see in response to `$PSVersionTable.PSVersion.ToString()` in a PowerShell console, for example: `7.0.0-rc.2`

PR #406, opened 2020-02-19
Tolerate read-only managed dependencies files

### LINKED ISSUE #400: Managed Dependencies: Gracefully handle the situation when the file storage is read-only
On the first function load request, PowerShell worker looks for the latest acceptable dependencies snapshot under D:\home\data\ManagedDependencies, and tries to update the timestamp on a special file (`.used`) within the snapshot folder. In certain scenarios, this file storage becomes temporarily read-only. Unfortunately, PowerShell worker treats a failure to update the timestamp as a failure to load the function, which makes the app non-functional until the file storage becomes writable again _and_ the worker is restarted.

There is not much we can do when the storage is read-only _before the initial managed dependencies installation_. However, we can make PowerShell worker _tolerate_ these errors _after the first acceptable snapshot is installed_. (The only reason to update the timestamp on the `.used` file is to prevent this snapshot from being purged, but they will not be purged from a read-only file storage anyway, so ignoring these errors is safe.)

AnatoliB:
Resolves #400:
- When updating the heartbeat (the timestamp on the .used file under a snapshot) fails because the file is read-only, log a warning but proceed.
- Make sure the worker does not purge the currently used snapshot even if the heartbeat has not been updated for long time.

We'll need this fix in `v2.x`, `v3.x/ps6`, and `v3.x/ps7` branches.

### DIFF
--- a/src/resources/PowerShellWorkerStrings.resx
+++ b/src/resources/PowerShellWorkerStrings.resx
@@ -257,7 +257,7 @@
     <value>Updating dependencies folder heartbeat for '{0}''.</value>
   </data>
   <data name="FailedToUpdateManagedDependencySnapshotHeartbeat" xml:space="preserve">
-    <value>Failed to update dependencies folder heartbeat for '{0}'. Exception {1}: '{2}''.</value>
+    <value>Failed to update dependencies folder heartbeat for '{0}'. Exception {1}: '{2}'.</value>
   </data>
   <data name="FailedToInstallDependenciesSnapshot" xml:space="preserve">
     <value>Failed to install dependencies into '{0}' (installation mode: {1}), removing the folder.</value>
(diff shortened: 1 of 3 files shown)


src/resources/PowerShellWorkerStrings.resx
```diff
@@ -256,6 +256,9 @@
   <data name="UpdatingManagedDependencySnapshotHeartbeat" xml:space="preserve">
     <value>Updating dependencies folder heartbeat for '{0}''.</value>
   </data>
+  <data name="FailedToUpdateManagedDependencySnapshotHeartbeat" xml:space="preserve">
+    <value>Failed to update dependencies folder heartbeat for '{0}'. Exception {1}: '{2}''.</value>
```
Francisco-Gamino: You have an extra single quote after '{2}''.

Francisco-Gamino (approved):
One minor comment; otherwise, LGTM.

src/resources/PowerShellWorkerStrings.resx
AnatoliB: Yes, thank you.

PR #413, opened 2020-02-28
Extract durable code to a separate folder, add unit tests

AnatoliB:
This refactoring has two goals:
- extract all the durable-related code to a separate folder/namespace;
- cover durable-related code with unit tests.

No _intended_ behavior changes.

### DIFF
--- a/src/Durable/DurableController.cs
+++ b/src/Durable/DurableController.cs
@@ -1,0 +1,140 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Collections.ObjectModel;
+    using System.Linq;
+    using System.Management.Automation;
+
+    using Newtonsoft.Json;
+    using WebJobs.Script.Grpc.Messages;
+
+    using PowerShellWorker.Utility;
+
+    internal class DurableController
+    {
+        private readonly bool _durableEnabled;
+        private readonly DurableFunctionInfo _durableFunctionInfo;
+        private readonly IPowerShellServices _powerShellServices;
+        private readonly IOrchestrationInvoker _orchestrationInvoker;
+        private OrchestrationBindingInfo _orchestrationBindingInfo;
+
+        public DurableController(
+            DurableFunctionInfo durableDurableFunctionInfo,
+            PowerShell pwsh)
+            : this(
+                Utils.AreDurableFunctionsEnabled(),
+                durableDurableFunctionInfo,
+                new PowerShellServices(pwsh),
+                new OrchestrationInvoker())
+        {
+        }
+
+        internal DurableController(
+            bool durableEnabled,
+            DurableFunctionInfo durableDurableFunctionInfo,
+            IPowerShellServices powerShellServices,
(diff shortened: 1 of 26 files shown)


src/Durable/DurableController.cs
```diff
@@ -0,0 +1,143 @@
...
+            _orchestrationInvoker = orchestrationInvoker;
+        }
+
+        public void BeforeFunctionInvocation(IList<ParameterBinding> inputData)
+        {
+            // If the function is an orchestration client, then we set the OrchestrationClient
+            // in the module context for the 'Start-NewOrchestration' function to use.
+            if (_durableFunctionInfo.IsOrchestrationClient)
+            {
+                ThrowIfDurableNotEnabled();
```
Francisco-Gamino, line 60: Does this method get called for non durable functions? If the answer is no, then would it be possible to move `ThrowIfDurableNotEnabled();` outside the branches? e.g., 

```
public void BeforeFunctionInvocation(IList<ParameterBinding> inputData)
{
   ThrowIfDurableNotEnabled();
   ...
```

AnatoliB: It actually does get called for all functions, including non-durable. I'm intentionally moving checking function type, checking the feature flag, and throwing the exception into the Durable folder/namespace. This logic is small and simple, but it already belongs to Durable.

src/Durable/DurableController.cs
```diff
@@ -0,0 +1,143 @@
...
+            // Quote from https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-bindings:
+            //
+            // "Orchestrator functions should never use any input or output bindings other than the orchestration trigger binding.
+            //  Doing so has the potential to cause problems with the Durable Task extension because those bindings may not obey the single-threading and I/O rules."
+            //
+            // Therefore, it's by design that input data contains only one item, which is the metadata of the orchestration context.
+            var context = inputData[0];
+
+            return new OrchestrationBindingInfo(
+                context.Name,
```
Francisco-Gamino, line 131: Is it possible that `content` is null at some point?

src/Durable/PowerShellServices.cs
```diff
@@ -0,0 +1,71 @@
...
+        private bool _hasSetOrchestrationContext = false;
+
+        public PowerShellServices(PowerShell pwsh)
+        {
+            _pwsh = pwsh;
+        }
+
+        public void SetOrchestrationClient(object orchestrationClient)
+        {
+            _pwsh.AddCommand("Microsoft.Azure.Functions.PowerShellWorker\\Set-FunctionInvocationContext")
```
Francisco-Gamino: Should `"Microsoft.Azure.Functions.PowerShellWorker\\Set-FunctionInvocationContext"` this a constant?

Francisco-Gamino (approved):
LGTM. Just a couple of minor comments. 

Are you planing to add end to end tests?

AnatoliB:
> Are you planing to add end to end tests?

There is one already: DurableEndToEndTests. I will add more when implementing major features.

src/Durable/PowerShellServices.cs
AnatoliB: Thank you, done.

src/Durable/DurableController.cs
AnatoliB: The only way for this to happen is if the Durable extension is broken so that it does not send the orchestration context data to the orchestrator function. In this case, many things will break, so this will be noticed before this line of code is executed. This cannot be caused by a user error.

PR #424, opened 2020-03-09
Use int.MaxValue for gRPC message size limit

### LINKED ISSUE #423: Increase gRpc channel message size to int.MaxValue
Related issue https://github.com/Azure/azure-functions-host/issues/5650

Channel Message Size is set here: https://github.com/Azure/azure-functions-powershell-worker/blob/8cab08b1440afb775a2375ebf1c54785387d5387/src/Messaging/MessagingStream.cs#L22-L26

Need to set these values to int.MaxValue and ignore host settings.

AnatoliB:
Use int.MaxValue for gRPC message size limit instead of grpcMaxMessageLength command line argument.

Resolves #423


pragnagopa (approved):
LGTM!

PR #427, opened 2020-03-13
[Durable Functions] Add fan-out/fan-in pattern support

AnatoliB:
See https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-overview#fan-in-out for the general information on the fan-out/fan-in pattern.

The main design idea:
- `Invoke-ActivityFunction` gets a new switch: `-NoWait`. When this switch is provided, `Invoke-ActivityFunction` returns a special "task" object and does _not_ stop the orchestrator.
- The orchestrator can call `Invoke-ActivityFunction` multiple times, and all these activities will be scheduled to run in parallel (subject to the regular concurrency constraints).
- The "task" objects returned by `Invoke-ActivityFunction` can be passed to the new `Wait-ActivityFunction` cmdlet. `Wait-ActivityFunction` checks if all the "tasks" are completed. If they are, it immediately returns the results from all of them and lets the orchestrator continue. If any of the "tasks" is not completed, it interrupts the orchestrator (in the same way as `Invoke-ActivityFunction` without `-NoWait` does).


docs/durable-experimental-instructions.md
```diff
@@ -2,36 +2,98 @@
...
+``` xml
+<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="1.8.3" />
+```
+
+## 3. App settings
 
-`PSWorkerEnableExperimentalDurableFunctions` = `"true"`
+Set the following app settings (if running on Azure) or just use the sample local.settings.json (if running locally):
 
-## 4. Configure AzureWebJobsStorage
```
Francisco-Gamino, line 31: It would be great to have a comment comment on what the key value pairs mean. Alternately, please add a link to the documentation for the settings.

examples/durable/DurableApp/FanOutFanInOrchestrator/run.ps1
```diff
@@ -0,0 +1,16 @@
+using namespace System.Net
+
+param($Context)
+
+Write-Host 'FanOutFanInOrchestrator: started.'
+
+$parallelTasks =
+    foreach ($Name in 'Tokyo', 'Seattle', 'London') {
```
Francisco-Gamino, line 8: This is just personal preference, but I feel that defining the array makes it easier to read. However, I am ok with either option.

```powershell

 foreach ($Name in @('Tokyo', 'Seattle', 'London')) {...}

```

src/Durable/WaitActivityFunctionCommand.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        [Parameter(Mandatory = true)]
+        [ValidateNotNull]
+        public ActivityInvocationTask[] Task { get; set; }
+
+        private readonly ActivityInvocationTracker _activityInvocationTracker = new ActivityInvocationTracker();
+
+        protected override void EndProcessing()
+        {
+            var privateData = (Hashtable)MyInvocation.MyCommand.Module.PrivateData;
```
Francisco-Gamino, line 24: Could `privateData` be null? If this is the case, please add check against null.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psd1
```diff
@@ -48,10 +48,18 @@ FormatsToProcess = @()
...
 
 # Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
-CmdletsToExport = @('Push-OutputBinding', 'Get-OutputBinding', 'Trace-PipelineObject', 'Set-FunctionInvocationContext', 'Invoke-ActivityFunction')
+CmdletsToExport = @(
+    'Push-OutputBinding',
+    'Get-OutputBinding',
+    'Trace-PipelineObject',
+    'Set-FunctionInvocationContext',
+    'Invoke-ActivityFunction',
+    'Wait-ActivityFunction')
```
Francisco-Gamino, line 62: Since we are adding the `Wait-ActivityFunction` cmdlet, we should probably increase the module version.

test/Unit/Durable/ActivityInvocationTrackerTests.cs
```diff
@@ -20,101 +21,209 @@ public class ActivityInvocationTrackerTests
...
+        [InlineData(true, true)]
+        public void WaitForActivityTasks_OutputsActivityResults_WhenAllTasksCompleted(
+            bool scheduled, bool completed)
+        {
+            var history = new[]
             {
-                Assert.True(stopwatch.ElapsedMilliseconds < delayBeforeStopping.TotalMilliseconds * 0.2);
-            }
+                CreateHistory("FunctionA", scheduled: scheduled, completed: completed, output: "\"Result1\""),
+                CreateHistory("FunctionA", scheduled: scheduled, completed: completed, output: "\"Result2\""),
```
Francisco-Gamino, line 143: In the `CreateHistory` we have two `FunctionA`. What's the purpose behind this? I apologize if this is supposed to be obvious. I see that below we check the outputs order which should be "Result1", "Result2", "Result3" which makes sense to me.

Francisco-Gamino (approved):
A few minor comments. Other than that, LGTM.

docs/durable-experimental-instructions.md
AnatoliB: Done.

examples/durable/DurableApp/FanOutFanInOrchestrator/run.ps1
AnatoliB: This does look like a personal preference, and I'm sure someone would prefer the less verbose style, so let me leave it as is.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psd1
AnatoliB: Done.

src/Durable/WaitActivityFunctionCommand.cs
AnatoliB: PowerShell promises it will contain at least an empty hashtable.

test/Unit/Durable/ActivityInvocationTrackerTests.cs
AnatoliB: This is actually important, but the purpose is not immediately obvious, you are right. I'm adding a comment.

examples/durable/DurableApp/FanOutFanInOrchestrator/run.ps1
Francisco-Gamino: Sounds good.

test/Unit/Durable/ActivityInvocationTrackerTests.cs
Francisco-Gamino: Great, thank you!

PR #441, opened 2020-03-24
Reload environment variables on FunctionEnvironmentReloadRequest

AnatoliB:
We need this to enable placeholders for PowerShell functions:
- For placeholder purposes, the Functions Host starts language workers with a generic set of environment variables.
- During specialization, the Functions Host sends a set of environment variables specific for the user app to the language worker, so that it has an opportunity to configure these variables without the need to restart.

### DIFF
--- a/src/RequestProcessor.cs
+++ b/src/RequestProcessor.cs
@@ -344,6 +344,19 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
 
         internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessage request)
         {
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+
+            var environmentReloadRequest = request.FunctionEnvironmentReloadRequest;
+            foreach (var (name, value) in environmentReloadRequest.EnvironmentVariables)
+            {
+                Environment.SetEnvironmentVariable(name, value);
+            }
+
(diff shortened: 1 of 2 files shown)


src/RequestProcessor.cs
```diff
@@ -344,6 +344,19 @@ internal StreamingMessage ProcessInvocationCancelRequest(StreamingMessage reques
 
         internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessage request)
         {
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+
+            var environmentReloadRequest = request.FunctionEnvironmentReloadRequest;
+            foreach (var (name, value) in environmentReloadRequest.EnvironmentVariables)
+            {
+                Environment.SetEnvironmentVariable(name, value);
```
Francisco-Gamino, line 353: We only want to set the variables for the current process, is that correct?

Francisco-Gamino (approved):
Just one comment (clarification). Other than that, LGTM.

src/RequestProcessor.cs
AnatoliB: Yes. Each worker will receive its own reload request and will take care of itself only.

Francisco-Gamino: Thanks for the clarification.

PR #445, opened 2020-04-08
[Durable Functions] Enable Durable Functions v1

AnatoliB:
Enable Durable Functions v1 support, most likely just temporarily. We want to do it now because DurableTask v2.* is not included in the Functions extension bundles yet, which complicates authoring templates and makes trying this out more difficult. We still have no intention to officially commit to Durable Functions v1 support, and we will probably disable it again when DurableTask v2.* is in the extension bundles.

/cc @anthonychu

### DIFF
--- a/src/Durable/DurableBindings.cs
+++ b/src/Durable/DurableBindings.cs
@@ -13,9 +13,13 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
         private const string OrchestrationTrigger = "orchestrationTrigger";
         private const string ActivityTrigger = "activityTrigger";
 
+        // For Durable v1 only
+        private const string OrchestrationClient = "orchestrationClient";
+
         public static bool IsDurableClient(string bindingType)
         {
-            return string.Compare(bindingType, DurableClient, StringComparison.OrdinalIgnoreCase) == 0;
+            return string.Compare(bindingType, DurableClient, StringComparison.OrdinalIgnoreCase) == 0
+                   || string.Compare(bindingType, OrchestrationClient, StringComparison.OrdinalIgnoreCase) == 0;
         }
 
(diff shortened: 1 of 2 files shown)


anthonychu:
Thanks for doing this. We can remove v1 support once we default to bundles v2 in our tooling (ETA summer).

anthonychu:
Actually we can probably remove support for `orchestrationClient` sooner, once this change is widely available in bundles v1: https://github.com/Azure/azure-functions-durable-extension/pull/1295 Still probably a couple months away.

src/Durable/DurableBindings.cs
```diff
@@ -13,9 +13,13 @@ internal static class DurableBindings
...
         private const string ActivityTrigger = "activityTrigger";
 
+        // For Durable v1 only
+        private const string OrchestrationClient = "orchestrationClient";
+
         public static bool IsDurableClient(string bindingType)
         {
-            return string.Compare(bindingType, DurableClient, StringComparison.OrdinalIgnoreCase) == 0;
+            return string.Compare(bindingType, DurableClient, StringComparison.OrdinalIgnoreCase) == 0
+                   || string.Compare(bindingType, OrchestrationClient, StringComparison.OrdinalIgnoreCase) == 0;
```
Francisco-Gamino, line 22: Do we log anywhere what Durable version is the customer using? If this information could be useful for debugging (or a CRI investigation), please add it. Otherwise, let's keep it as is.

Francisco-Gamino (approved):
One minor comment; otherwise, LGTM.

src/Durable/DurableBindings.cs
AnatoliB: I agree this would be useful, but let's do it separately: #446 filed

PR #449, opened 2020-04-08
Upgrade .NET SDK to 2.2.110

AnatoliB:
Upgrade to the latest **LTS** .NET Core 2.2 SDK

### DIFF
--- a/tools/helper.psm1
+++ b/tools/helper.psm1
@@ -40,7 +40,7 @@ function Test-DotnetSDK
 
     if (Test-Path $dotnetExePath) {
         $installedVersion = & $dotnetExePath --version
-        return $installedVersion -ge $MinimalSDKVersion
+        return [version]$installedVersion -ge [version]$MinimalSDKVersion
     }
     return $false
 }
--- a/tools/helper.psm1
+++ b/tools/helper.psm1
@@ -35,12 +35,20 @@ function Find-Dotnet
     }
 }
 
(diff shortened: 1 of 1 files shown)


tools/helper.psm1
```diff
@@ -7,7 +7,8 @@ using namespace System.Runtime.InteropServices
 
 $IsWindowsEnv = [RuntimeInformation]::IsOSPlatform([OSPlatform]::Windows)
 $RepoRoot = (Resolve-Path "$PSScriptRoot/..").Path
-$MinimalSDKVersion = '2.2.0'
+$MinimalSDKVersion = '2.2.110'
```
Francisco-Gamino, line 10: Question: For .NET `v2.2.8`, based on the guidance here: https://dotnet.microsoft.com/download/dotnet-core/2.2, we could also install `2.2.207`, is that correct?

AnatoliB: `2.2.207` is the latest for **VS2019**, but `2.2.110` is the latest for **VS2017**. In the v2.x branch, we are still using **VS2017**. (We could upgrade to **VS2019** like we did with v3.x, but I'm a bit reluctant to do this for a legacy branch, unless there is a strong reason.)

Francisco-Gamino: Ok, let's keep it `2.2.110` and continue to use **VS2017**.

Francisco-Gamino (approved):
LGTM

PR #450, opened 2020-04-15
[Durable Functions] Improve usage error handling

AnatoliB:
Throw exceptions with user-friendly messages when trying to invoke a non-existing or an incorrect activity function


Francisco-Gamino (approved):
LGTM

PR #480, opened 2020-06-25
Add a basic monitor sample to DurableApp

michaelpeng36:
This adds a tentative template for the monitoring pattern with placeholder activity functions "GetJobStatus" and "SendAlert" to the DurableApp.

### DIFF
--- a/examples/durable/DurableApp/GetJobStatus/run.ps1
+++ b/examples/durable/DurableApp/GetJobStatus/run.ps1
@@ -1,8 +1,9 @@
 param($name)
 
-if ((Get-Date).ToUniversalTime() -lt $name) {
-    Write-Host "Not Completed"
+$randomNumber = Get-Random -Maximum 100
+if ($randomNumber -lt 60) {
+    "Not Completed"
 }
 else {
-    Write-Host "Completed"
+    "Completed"
 }
--- a/examples/durable/DurableApp/GetJobStatus/run.ps1
+++ b/examples/durable/DurableApp/GetJobStatus/run.ps1
@@ -1,0 +1,8 @@
+param($name)
+
+if ((Get-Date).ToUniversalTime() -lt $name) {
+    Write-Host "Not Completed"
+}
+else {
+    Write-Host "Completed"
+}
--- a/examples/durable/DurableApp/MonitorOrchestrator/run.ps1
+++ b/examples/durable/DurableApp/MonitorOrchestrator/run.ps1
@@ -1,5 +1,3 @@
-using namespace System.Net
-
 param($Context)
 
 Write-Host 'MonitorOrchestrator: started.'
@@ -12,12 +10,12 @@ while ($Context.CurrentUtcDateTime -lt $expiryTime) {
     $jobStatus = Invoke-ActivityFunction -FunctionName 'GetJobStatus' -Input $jobId
     if ($jobStatus -eq "Completed") {
         # Perform an action when a condition is met.
-        $output = Invoke-ActivityFunction -FunctionName 'SendAlert' -Input $machineId
+        $output = Invoke-ActivityFunction -FunctionName 'SendAlert' -Input "foo"
         break
     }
 
     # Orchestration sleeps until this time.
-    Start-Sleep -Seconds $pollingInterval
+    Start-DurableTimer -Seconds $pollingInterval
 }
 
 # Perform more work here, or let the orchestration end.
--- a/examples/durable/DurableApp/MonitorOrchestrator/run.ps1
+++ b/examples/durable/DurableApp/MonitorOrchestrator/run.ps1
@@ -3,6 +3,7 @@ param($Context)
 Write-Host 'MonitorOrchestrator: started.'
 
 $jobId = $Context.Input.JobId
+$machineId = $Context.Input.MachineId
 $pollingInterval = $Context.Input.PollingInterval
 $expiryTime = $Context.Input.ExpiryTime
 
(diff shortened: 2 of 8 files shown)


examples/durable/DurableApp/MonitorOrchestrator/run.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+$jobId = $Context.Input.JobId
+$pollingInterval = $Context.Input.PollingInterval
+$expiryTime = $Context.Input.ExpiryTime
+
+while ($Context.CurrentUtcDateTime -lt $expiryTime) {
+    $jobStatus = Invoke-ActivityFunction -FunctionName 'GetJobStatus' -Input $jobId
+    if ($jobStatus -eq "Completed") {
+        # Perform an action when a condition is met.
+        $output = Invoke-ActivityFunction -FunctionName 'SendAlert' -Input $machineId
```
AnatoliB, line 14: Where is `$machineId` from?

examples/durable/DurableApp/GetJobStatus/run.ps1
```diff
@@ -0,0 +1,8 @@
+param($name)
+
+if ((Get-Date).ToUniversalTime() -lt $name) {
```
AnatoliB: This looks a bit strange: you are comparing the current time with `$name`. Though technically legal, the intent is not clear. If you were going to encode the "desired" completion date in the name, you probably need to construct and pass this value in the MonitorStart code (I can see you are just passing `$null` there now), and then either give the parameter a better name or describe this in comments. Alternatively, consider returning "Not Completed" or "Completed" randomly, perhaps with 90/10% probability.

examples/durable/DurableApp/MonitorStart/run.ps1
```diff
@@ -0,0 +1,15 @@
+using namespace System.Net
+
+param($Request, $TriggerMetadata)
+
+Write-Host 'MonitorStart started'
+
+$OrchestratorInputs = @{ JobId = $null; PollingInterval = 30; ExpiryTime = (Get-Time).addSeconds(60)}
```
AnatoliB: You probably want to pass a UTC date/time as ExpiryTime.

examples/durable/DurableApp/MonitorOrchestrator/run.ps1
```diff
@@ -0,0 +1,26 @@
+using namespace System.Net
```
AnatoliB: This line is probably unnecessary.

examples/durable/DurableApp/MonitorStart/run.ps1
```diff
@@ -0,0 +1,15 @@
+using namespace System.Net
```
AnatoliB: This line is probably unnecessary.

AnatoliB (changes requested):
See the comments.

examples/durable/DurableApp/MonitorOrchestrator/run.ps1
michaelpeng36: I believe the user supplies this value, as it is not passed into the orchestrator in either the C# or JavaScript template.

examples/durable/DurableApp/MonitorOrchestrator/run.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+$jobId = $Context.Input.JobId
+$pollingInterval = $Context.Input.PollingInterval
+$expiryTime = $Context.Input.ExpiryTime
+
+while ($Context.CurrentUtcDateTime -lt $expiryTime) {
+    $jobStatus = Invoke-ActivityFunction -FunctionName 'GetJobStatus' -Input $jobId
+    if ($jobStatus -eq "Completed") {
+        # Perform an action when a condition is met.
+        $output = Invoke-ActivityFunction -FunctionName 'SendAlert' -Input "foo"
```
AnatoliB: Let's come up with a more meaningful and realistic sample value than "foo". Perhaps passing $jobId there makes sense?

examples/durable/DurableApp/MonitorStart/run.ps1
```diff
@@ -0,0 +1,13 @@
+param($Request, $TriggerMetadata)
+
+Write-Host 'MonitorStart started'
+
+$OrchestratorInputs = @{ JobId = $null; PollingInterval = 10; ExpiryTime = (Get-Date).addSeconds(60).ToUniversalTime() }
```
AnatoliB: You need to convert to UTC first, and then add the 60 seconds. The order matters: local time is ambiguous and not continuous, so the results of arithmetic operations can be surprising (for example, on the nights of switching to/from DST).

AnatoliB (changes requested):
A couple of minor comments, almost done.

examples/durable/DurableApp/MonitorOrchestrator/run.ps1
michaelpeng36: $jobId makes sense, but I believe $machineId might be better to maintain parallelism with the C#, JavaScript, and Python templates.

AnatoliB: The first priority here is to make it a working self-sufficient sample. If you want to pass `$machineId`, it should be initialized first, otherwise `$null` will be passed. Or pass a hardcoded value like `"machine id"`. `"foo"` is just strange: as a reader of this code, I have no idea how to interpret it.

Making this code consistent with samples for other languages is nice but completely optional here. This consistency would be important on a documentation page where multiple samples are presented together on different tab, but this is not where we are.

AnatoliB (approved):

PR #486, opened 2020-07-09
Implement CurrentUtcDateTime property for Durable functions

michaelpeng36:
Added the CurrentUtcDateTime instance property to OrchestrationContext for use in the monitoring pattern. Added unit tests for the CurrentUtcDateTime feature (assumptions made about history structure are documented above the respective test).


michaelpeng36:
From my local experiments with pre-existing patterns (FunctionChaining and FanInFanOut), it appears that the history always contains an OrchestratorStarted and ExecutionStarted event whenever OrchestrationInvoker.Invoke() is called. Should adjustments to OrchestrationInvokerTests.cs be made?

src/Durable/OrchestrationInvoker.cs
```diff
@@ -19,6 +20,16 @@ public Hashtable Invoke(OrchestrationBindingInfo orchestrationBindingInfo, IPowe
             try
             {
                 var outputBuffer = new PSDataCollection<object>();
+                var context = orchestrationBindingInfo.Context;
+
+                // context.History should never be null when initializing CurrentUtcDateTime
+                if (context.History != null)
```
AnatoliB: If context.History is not supposed to be null, let's remove this check and adjust the tests accordingly, as we discussed. If it ever _happens_ to be null, we should not hide this fact, let the code throw an exception.

src/Durable/OrchestrationInvoker.cs
```diff
@@ -19,6 +20,16 @@ public Hashtable Invoke(OrchestrationBindingInfo orchestrationBindingInfo, IPowe
...
+                {
+                    var orchestrationStart = context.History.FirstOrDefault(
```
AnatoliB: Similarly, if the history is always supposed to have the OrchestratorStarted event, let's invoke `First` instead of `FirstOrDefault`, so we get an exception right there if this contract is ever broken.

src/Durable/OrchestrationInvoker.cs
```diff
@@ -19,6 +20,16 @@ public Hashtable Invoke(OrchestrationBindingInfo orchestrationBindingInfo, IPowe
...
+                        (e) => e.EventType == HistoryEventType.OrchestratorStarted);
```
AnatoliB: The parentheses around `e` are unnecessary.

src/Durable/OrchestrationContext.cs
```diff
@@ -30,6 +31,9 @@ public class OrchestrationContext
         [DataMember]
         internal HistoryEvent[] History { get; set; }
 
+        [DataMember]
```
AnatoliB: We don't expect CurrentUtcDateTime to be passed to the PowerShell worker from the host, correct? As far as I understant, this property is initialized and used by the PowerShell worker only. In this case, the `DataMember` attribute is not needed, please remove it.

src/Durable/OrchestrationContext.cs
```diff
@@ -30,6 +31,9 @@ public class OrchestrationContext
...
+        public DateTime CurrentUtcDateTime {get; internal set; }
```
AnatoliB, line 34: Just a thought, no need to do anything about it right now:

We are doing something weird here. We are reusing the same class for two entirely different purposes. First of all, we use it for deserializing the data required to start the orchestrator, which is passed to the worker from the host serialized to Json, and this is why we the relevant properties are marked with `DataMember`. But then, we want to expose _some_ properties to the user script  (such as `Input`, `CurrentUtcDateTime`), while hiding other properties. We may want to separate these into two distinct classes, I just don't know how to easily do this right now. But, if you have a good idea, try that.

Again, we don't _have_ to do this right now, we will not block this PR because of that.

src/Durable/ActivityInvocationTracker.cs
```diff
@@ -75,6 +86,16 @@ internal class ActivityInvocationTracker
             var allTasksCompleted = completedEvents.Count == tasksToWaitFor.Count;
             if (allTasksCompleted)
             {
+                var newOrchestrationStart = context.History.FirstOrDefault(
+                    (e) => e.EventType == HistoryEventType.OrchestratorStarted &&
+                    e.Timestamp > context.CurrentUtcDateTime
```
AnatoliB: Please format this in a way that makes it clear that both `e.EventType == ...` and `e.Timestamp > ...` belong to the same lambda body. I would just insert a few spaces so that `e.Timestamp` is directly below `e.EventType`.

AnatoliB: Thinking about this timestamp comparison: in theory, if the function is executed by a super-fast computer, the timestamps of different events may coincide. Though this may sound quite improbable, it may eventually happen. What would be the consequences of that? I'm not sure whether we will have any negative consequences or not, please let me know if you see a problem here.

michaelpeng36: This will most likely lead to an error in the CurrentUtcDateTime during replays. For example suppose we have two OrchestratorStarted events OS1 and OS2 with the same timestamp T1. Let TS and TC denote task scheduled events and task completed events respectively, and suppose we have the first OrchestratorStarted event OS3 with timestamp T2 > T1. If the history resembles TS -> OS1 -> TC -> TS -> OS2 -> TC -> TS -> OS3 in chronological order, then upon replaying the first task completed, the CurrentUtcDateTime (representing the DateTime between OS1 and OS2) is updated to T2, when it should be T1.

michaelpeng36: I will add a check for this in the CurrentUtcDateTime tests. I believe an appropriate way to modify this check is to use the HistoryEvent.IsProcessed instance property instead of the Timestamp.

test/E2E/Azure.Functions.PowerShellWorker.E2E/Azure.Functions.PowerShellWorker.E2E/DurableEndToEndTests.cs
```diff
@@ -85,6 +85,94 @@ public async Task DurableClientFollowsAsyncPattern()
...
+
+            var initialResponseBody = await initialResponse.Content.ReadAsStringAsync();
+            dynamic initialResponseBodyObject = JsonConvert.DeserializeObject(initialResponseBody);
+            Assert.NotNull(initialResponseBodyObject.id);
+            var statusQueryGetUri = (string)initialResponseBodyObject.statusQueryGetUri;
+            Assert.Equal(location.ToString(), statusQueryGetUri);
+            Assert.NotNull(initialResponseBodyObject.sendEventPostUri);
+            Assert.NotNull(initialResponseBodyObject.purgeHistoryDeleteUri);
+            Assert.NotNull(initialResponseBodyObject.terminatePostUri);
+            Assert.NotNull(initialResponseBodyObject.rewindPostUri);
```
AnatoliB: I suggest removing all the asserts on lines 95-105. These verifications are performed by the DurableClientFollowsAsyncPattern test, and these is where they belong. In this test, focus on verifying what is relevant specifically for CurrrentUtcDateTime - this will make the test easier to understand, and the potential issues easier to isolate.

AnatoliB (changes requested):

AnatoliB (approved):
Looks good to me. Please tell me if you are done (ignoring the NuGetCommand step failure), so we can merge it.

PR #488, opened 2020-07-13
Adding the DurableTimer feature as well as E2E and unit tests

michaelpeng36:
Adds the Start-DurableTimer cmdlet, which is similar to Invoke-ActivityFunction in that it restarts the orchestrator; however it does so only after a specified amount of time. Upon reaching a TimerFired event, the DurableTimer should update the OrchestrationContext's CurrentUtcDateTime property to the Timestamp of the next OrchestratorStarted event.


src/Durable/DurableTimerTask.cs
```diff
@@ -0,0 +1,51 @@
...
+            bool isCompleted,
+            bool isCanceled,
+            CreateDurableTimerAction createDurableTimerAction)
+            {
+                IsCompleted = isCompleted;
+                IsCanceled = isCanceled;
+                _createDurableTimerAction = createDurableTimerAction;
+            }
+
+        public void Stop()
```
AnatoliB: Do we need this method? How would it be used?

src/Durable/DurableTimerTask.cs
```diff
@@ -0,0 +1,51 @@
...
+
+#pragma warning disable 1591 // "Mixing XML comments for publicly visible type"
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System;
+    public class DurableTimerTask
+    {
+        public bool IsCompleted { get; }
+        public bool IsCanceled { get; set; }
```
AnatoliB: Do we need these to be public?

src/Durable/DurableTimerTask.cs
```diff
@@ -0,0 +1,51 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#pragma warning disable 1591 // "Mixing XML comments for publicly visible type"
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System;
+    public class DurableTimerTask
```
AnatoliB: In fact, do we need this entire class? I don't see any references to it.

src/Durable/CreateDurableTimerAction.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        /// <summary>
+        /// The DateTime at which the timer will fire.
+        /// </summary>
+        public readonly DateTime FireAt;
+        
+        /// <summary>
+        /// Whether the Durable Timer was canceled.
+        /// </summary>
+        public bool IsCanceled;
```
AnatoliB: readonly?

src/Durable/DurableTimer.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    using System;
+    using System.Linq;
+    using System.Threading;
+
+    internal class DurableTimer
+    {
+        private readonly ManualResetEvent _waitForStop = new ManualResetEvent(initialState: false);
+
+        public void CreateTimerAndStop_OrContinue(OrchestrationContext context, DateTime fireAt)
```
AnatoliB: Let's remove the underscore from the method name to stay consistent with the de facto naming convention. We make an exception for unit test names only.

src/Durable/DurableTimer.cs
```diff
@@ -0,0 +1,79 @@
...
+                var newOrchestrationStart = context.History.FirstOrDefault(
+                        e => e.EventType == HistoryEventType.OrchestratorStarted &&
+                            !e.IsProcessed
+                    );
+                
+                if (newOrchestrationStart != null)
+                {
+                    context.CurrentUtcDateTime = newOrchestrationStart.Timestamp.ToUniversalTime();
+                    newOrchestrationStart.IsProcessed = true;
+                }
```
AnatoliB: We have this piece of code (lines 30-39) duplicated multiple times now (once here and twice in `ActivityInvocationTracker`). I think it deserves to be extracted to a method, perhaps on a new class (something like `CurrentUtcDateTimeUpdater`).

src/Durable/CreateDurableTimerAction.cs
AnatoliB: Do you know what this property is for? I don't see it used anywhere.

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,260 @@
...
+    using System.Linq;
+    using System.Management.Automation;
+    using System.Threading;
+    using Microsoft.Azure.Functions.PowerShellWorker.Durable;
+    using Moq;
+    using Xunit;
+
+    public class DurableTimerTests
+    {
+        private const DateTimeKind utc = DateTimeKind.Utc;
```
AnatoliB: I don't think this constant serves any useful purpose, let's use DateTimeKind.Utc directly.

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        private DateTime time = new DateTime(2020, 1, 1, 0, 0, 0, 0, utc);
```
AnatoliB: Let's come up with a better name: `time` is just too generic and does not give you any clue about the intent. `startTime`?

AnatoliB: Also, naming convention: we prefix non-constant private fields with `_`.

AnatoliB: Also, I think it can be `readonly`.

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        private int _nextEventId = 1;
+
+        // Verifies that CreateTimer waits for the time elapsed 
+        [Theory]
+        [InlineData(true, false)]
+        [InlineData(false, true)]
+        public void CreateTimerOrContinue_WaitsUntilTimerFires(bool timerScheduledAndFired, bool expectedWaitForStop)
+        {
+            int durationSeconds = longIntervalSeconds;
+            DateTime startTime = time;
```
AnatoliB: If it never changes, I suggest you just use the field directly, no need to create a copy.

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,260 @@
...
+    using System.Threading;
+    using Microsoft.Azure.Functions.PowerShellWorker.Durable;
+    using Moq;
+    using Xunit;
+
+    public class DurableTimerTests
+    {
+        private const DateTimeKind utc = DateTimeKind.Utc;
+        private DateTime time = new DateTime(2020, 1, 1, 0, 0, 0, 0, utc);
+        private int longIntervalSeconds = 5;
```
AnatoliB: `const`?

AnatoliB: Naming convention for constants: the first letter is uppercased, e.g. `LongIntervalSeconds`.

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        {
+            return histories.Aggregate((a, b) => a.Concat(b).ToArray());
+        }
+
+        private int GetUniqueEventId()
+        {
+            return _nextEventId++;
+        }
+
+        private Hashtable InvokeOrchestration(bool completed, PSDataCollection<object> output = null)
```
AnatoliB: Not used here?

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,260 @@
...
+                    EventId = orchestratorStartEventId,
+                    Timestamp = date,
+                    IsProcessed = isProcessed
+                }
+            );
+
+            return history.ToArray();
+        }
+
+        private static HistoryEvent[] MergeHistories(params HistoryEvent[][] histories)
```
AnatoliB: We have multiple copies of some methods like this across multiple Durable unit test classes. Please extract them to a class (something like `DurableTestUtilities` - or invent a better name if you can).

AnatoliB (changes requested):
Please address the comments.

test/Unit/Durable/DurableTimerTests.cs
AnatoliB: Consider just invoking `DateTime.UtcNow` (otherwise this code makes me think the hardcoded date has some special meaning).

src/Durable/DurableTimer.cs
michaelpeng36: Sounds good. My main concern is that the grouping of terms might be ambiguous, i.e. CreateTimer and (Stop or Continue) vs. (CreateTimerAndStop) or Continue. I am changing the name to StopAndCreateTimerOrContinue. Let me know if you see a similar issue in this name.

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+
+    public class DurableTimerTests
+    {
+        private const int LongIntervalSeconds = 5;
+        private const int ShortIntervalSeconds = 3;
+        private static readonly DateTime startTime = DateTime.UtcNow;
+        private readonly DateTime fireAt = startTime.AddSeconds(ShortIntervalSeconds);
+        private static readonly DateTime restartTime = startTime.AddSeconds(LongIntervalSeconds);
+        private static readonly DateTime shouldNotHitTime = restartTime.AddSeconds(LongIntervalSeconds);
+        private OrchestrationInvoker _orchestrationInvoker = new OrchestrationInvoker();
```
AnatoliB: Not used?

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        private OrchestrationBindingInfo _orchestrationBindingInfo = new OrchestrationBindingInfo(
```
AnatoliB: Not used?

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+                                                                            "ContextParameterName",
+                                                                            new OrchestrationContext());
+        private Mock<IPowerShellServices> _mockPowerShellServices = new Mock<IPowerShellServices>();
```
AnatoliB: Not used?

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+    using System.Threading;
+    using Microsoft.Azure.Functions.PowerShellWorker.Durable;
+    using Moq;
+    using Xunit;
+
+    public class DurableTimerTests
+    {
+        private const int LongIntervalSeconds = 5;
+        private const int ShortIntervalSeconds = 3;
+        private static readonly DateTime startTime = DateTime.UtcNow;
```
AnatoliB: Naming convention for non-const data members: prefix with underscore (e.g. `_startTime`)

test/Unit/Durable/DurableTestUtilities.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
AnatoliB: Please copy the header from other files:

```
//
// Copyright (c) Microsoft. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.
//
```

test/Unit/Durable/DurableTimerTests.cs
michaelpeng36: Oops it looks like I missed these after I removed the unnecessary methods.

examples/durable/DurableApp/MonitorOrchestrator/run.ps1
```diff
@@ -8,7 +8,7 @@ $output = @()
 
 $jobId = $Context.Input.JobId
 $machineId = $Context.Input.MachineId
-$pollingInterval = $Context.Input.PollingInterval
+$pollingInterval = New-Timespan -Seconds $Context.Input.PollingInterval
```
AnatoliB: Nit: should be `New-TimeSpan`

src/Durable/StartDurableTimerCommand.cs
```diff
@@ -22,15 +22,15 @@ public class StartDurableTimerCommand : PSCmdlet
         /// </summary>
         [Parameter(Mandatory = true)]
         [ValidateNotNullOrEmpty]
-        public int Seconds { get; set; }
+        public TimeSpan Duration { get; set; }
```
AnatoliB, line 25: We could in fact allow both `Seconds` and `Duration` (and make them mutually exclusive via PowerShell parameter sets). But this is a minor thing, let's leave it as is for now.

AnatoliB (approved):

PR #489, opened 2020-07-15
Simplify Durable Functions instructions and samples/tests configurations

AnatoliB:
Simplify Durable Functions instructions and samples/tests configurations:
- No need to pin the `Microsoft.Azure.WebJobs.Extensions.DurableTask` version anymore;
- No need to use the `AllowSynchronousIO` feature flag anymore.

/cc: @anthonychu


Francisco-Gamino (approved):
LGTM

PR #496, opened 2020-07-22
Fix Azure Pipeline builds

AnatoliB:
* Skip publishing NuGet on pull requests
* Use BuildNumber from Azure Pipelines instead of Appveyor


Francisco-Gamino:
LGTM

PR #499, opened 2020-07-23
Fix CurrentUtcDateTime E2E test

michaelpeng36:
Adjusted path separator


test/E2E/Azure.Functions.PowerShellWorker.E2E/Azure.Functions.PowerShellWorker.E2E/DurableEndToEndTests.cs
```diff
@@ -183,14 +188,14 @@ public async Task DurableExecutionReplaysCurrentUtcDateTime()
...
         {
-            if (indices.Capacity > 0)
+            if (indices.Length > 0)
             {
                 var expected = array[indices[0]];
-                for (int i = 1; i < indices.Capacity; i++)
+                for (int i = 1; i < indices.Length; i++)
                 {
-                    Assert.True(array.Length > indices[i], $"Array length is {array.Length} but index is {indices[i]}");
+                    Assert.True(indices[i] < array.Length, $"Array capacity is {array.Length} but index is {indices[i]}");
```
AnatoliB: Shouldn't the message say "Array length is..." instead of "Array capacity is..."?

michaelpeng36: Yes, I originally had an List<int> instead of an int[]. I will change it.

PR #512, opened 2020-08-03
Log a warning when a command is not found or a module is not found

### LINKED ISSUE #491: Issue an actionable warning on CommandNotFoundException
**Problem**
When users try to invoke a command that cannot be found in any modules, they get just the standard PowerShell error: `The term 'xxx' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.` Sometimes they do get into this situation because of a typo, but very often this happens because they are missing modules, or because these modules don't work in PowerShell Core. It takes time and effort to realize that, and then it takes time and effort to figure out _how_ to make the required modules available. Even though we have this [documented](https://docs.microsoft.com/azure/azure-functions/functions-reference-powershell#dependency-management), the documentation is not immediately discoverable, so new Function users spend more time and effort to invent their own ways, and  sometimes come up with inefficient and difficult to support solutions.

**Proposal**
- Detect this error during function execution and log a warning with suggestions and documentation links, something like: `The app may be missing a module containing the 'xxx' command definition. If this command belongs to a module available on the PowerShell gallery, add a reference to this module to requirements.psd1. Make sure this module is compatible with PowerShell Core 6 or PowerShell 7. For more details, see <doc link>`. /cc @eamonoreilly for the message wording.
- _Consider_ going further: query the PS Gallery automatically and come up with a line of code to be inserted into `requirements.psd1`, include this line into the warning text.

AnatoliB:
Resolves #491 

When a command is invoked from the function code, but this command cannot be found in any available module, logs the following warning:
> The app may be missing a module containing the '{0}' command definition. If this command belongs to a module available on the PowerShell Gallery, add a reference to this module to requirements.psd1. Make sure this module is compatible with PowerShell 7. For more details, see https://aka.ms/functions-powershell-managed-dependency.

When `Import-Module` is invoked from the function code but this module is not available, logs the following warning:
> The app may be missing the '{0}' module. If '{0}' is available on the PowerShell Gallery, add a reference to this module to requirements.psd1. Make sure this module is compatible with PowerShell 7. For more details, see https://aka.ms/functions-powershell-managed-dependency.

/cc @eamonoreilly @stefanushinardi for the warning message wording.

### DIFF
--- a/src/resources/PowerShellWorkerStrings.resx
+++ b/src/resources/PowerShellWorkerStrings.resx
@@ -322,4 +322,7 @@
   <data name="FunctionDoesNotHaveProperActivityFunctionBinding" xml:space="preserve">
     <value>Function '{0}' does not have an activityTrigger input binding.</value>
   </data>
+  <data name="CommandNotFoundUserWarning" xml:space="preserve">
+    <value>The app may be missing a module containing the '{0}' command definition. If this command belongs to a module available on the PowerShell gallery, add a reference to this module to requirements.psd1. Make sure this module is compatible with PowerShell 7. For more details, see https://aka.ms/functions-powershell-managed-dependency.</value>
+  </data>
 </root>
--- a/src/resources/PowerShellWorkerStrings.resx
+++ b/src/resources/PowerShellWorkerStrings.resx
@@ -311,7 +311,7 @@
     <value>CommandNotFoundException detected (exception).</value>
   </data>
   <data name="CommandNotFoundUserWarning" xml:space="preserve">
-    <value>The app may be missing a module containing the '{0}' command definition. If this command belongs to a module available on the PowerShell Gallery, add a reference to this module to requirements.psd1. Make sure this module is compatible with PowerShell 7. For more details, see https://aka.ms/functions-powershell-managed-dependency.</value>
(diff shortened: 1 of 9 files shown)


src/resources/PowerShellWorkerStrings.resx
```diff
@@ -310,6 +310,18 @@
   <data name="CommandNotFoundException_Exception" xml:space="preserve">
     <value>CommandNotFoundException detected (exception).</value>
   </data>
+  <data name="CommandNotFoundUserWarning" xml:space="preserve">
+    <value>The app may be missing a module containing the '{0}' command definition. If this command belongs to a module available on the PowerShell Gallery, add a reference to this module to requirements.psd1. Make sure this module is compatible with PowerShell 7. For more details, see https://aka.ms/functions-powershell-managed-dependency.</value>
```
Francisco-Gamino: Should we use `function app` instead of `app` for consistency? 

Also, instead of saying `PowerShell 7`, should we have a string that reference the latest version? E.g., `PowerShell {1}` where `{1}` is the current version, in this case 7.

Francisco-Gamino (approved):
One minor comment. Otherwise, LGTM

src/resources/PowerShellWorkerStrings.resx
AnatoliB: Agreed with "function app", fixed.
Regarding "PowerShell 7" - I'm not sure it's worth it, given the fact that the PowerShell team decided to never change the major version number, so it is highly unlikely we'll ever see "PowerShell 8" etc. If this happens, we can change this.

Francisco-Gamino: Sounds good. Thanks Anatoli.

PR #513, opened 2020-08-05
Log warnings and errors produced by Save-Module during managed dependency installation

### LINKED ISSUE #484: Insufficient logs when failing to install a module via Managed Dependencies
When a required module is mentioned in `requrements.psd1` and this module installation fails, PowerShell worker logs the error reported by the `Save-Module` cmdlet. However, this error message is often too generic: `Unable to save the module X`. This message is not helpful and not actionable. In the meantime, `Save-Module` may report more valuable messages to the Warning stream, but PowerShell worker does not log those, which makes the investigation difficult.

For example, the `AWSPowerShell.NetCore` module cannot be installed via Managed Dependencies on Consumption because it is too large to fit into D:\local\Temp. The error message we are getting now is still `Unable to save the module X`. The warning produced by `Save-Module` would point the user to the right direction: `Package 'AWSPowerShell.NetCore' failed to be installed because: There is not enough space on the disk. : 'D:\local\Temp\250686550\AWSPowerShell.NetCore.4.0.6.0\AWSPowerShell.NetCore.dll-Help.xml'`.

So, let's log warnings produced by `Save-Module`.

AnatoliB:
Resolves #484

### DIFF
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -48,12 +48,13 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
             IInstalledDependenciesLocator installedDependenciesLocator = null,
             IDependencySnapshotInstaller installer = null,
             INewerDependencySnapshotDetector newerSnapshotDetector = null,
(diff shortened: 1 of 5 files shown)


Francisco-Gamino (approved):
LGTM

PR #514, opened 2020-08-05
Added NoWait switch parameter to Start-DurableTimer cmdlet

michaelpeng36:
Added a NoWait switch parameter to the Start-DurableTimer cmdlet that will cause it to return a DurableTimerTask object. A DurableTimerTask contains basic information regarding the timeout task (whether it is completed, whether it has been canceled, and the time at which it will time out) and will be able to be canceled by the user.


src/Durable/DurableTimerTask.cs
```diff
@@ -0,0 +1,25 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#pragma warning disable 1591 // "Mixing XML comments for publicly visible type"
```
AnatoliB: "Mixing" is a typo, should be "Missing"

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -26,13 +26,13 @@ public class DurableTimerTests
 
         // Verifies that CreateTimer waits for the time elapsed 
         [Theory]
-        [InlineData(true, false)]
-        [InlineData(false, true)]
-        public void CreateTimerOrContinue_WaitsUntilTimerFires(bool timerScheduledAndFired, bool expectedWaitForStop)
+        [InlineData(true, true, false)]
+        [InlineData(false, false, true)]
```
AnatoliB, line 30: Do we need to add the (true, false, ...) case?

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -143,6 +204,13 @@ private void EmulateStop(DurableTimer durableTimer)
             durableTimer.Stop();
         }
 
+        private void VerifyCreateDurableTimerActionAdded(OrchestrationContext context, DateTime fireAt)
+        {
+            var actions = DurableTestUtilities.GetCollectedActions(context);
+            var action = (CreateDurableTimerAction)actions[actions.Count - 1];
```
AnatoliB: Use `.Last()` instead of `[Count - 1]`: this is simpler and communicates the intent better.

test/Unit/Durable/DurableTimerTests.cs
```diff
@@ -42,43 +42,101 @@ public void CreateTimerOrContinue_WaitsUntilTimerFires(bool timerScheduledAndFir
...
+
+            var context = new OrchestrationContext { History = history };
 
-            if (timerScheduledAndFired)
+            if (!timerCreated)
             {
+                EmulateStop(_durableTimer);
+            }
+
+            _durableTimer.StopAndCreateTimerOrContinue(context: context, fireAt: _fireAt, noWait: false, _ => { Assert.True(false, "Unexpected output"); } );
```
AnatoliB, line 119: Perhaps this is the only test that must have `Assert.True(false, "Unexpected output")`. Consider removing this assert from other tests, just pass `_ => { }`. First of all, this assert will not distract the reader of other tests. Also, if we break the production code and it suddenly start returning something when -NoWait is absent, we will have exactly one unit test failing, and the name of this failing test will point out the reason.

AnatoliB (changes requested):
A few minor comments

test/Unit/Durable/DurableTimerTests.cs
AnatoliB: Discussed offline: (true, false) does not seem to be a valid use case.

AnatoliB (approved):

PR #516, opened 2020-08-07
Added human interaction pattern template

michaelpeng36:
Added a template for implementing the human interaction pattern in the DurableApp folder


examples/durable/DurableApp/HumanInteractionOrchestrator/run.ps1
```diff
@@ -0,0 +1,28 @@
...
+
+param($Context)
+
+Write-Host 'HumanInteractionOrchestrator: started.'
+
+$output = @()
+
+$duration = $Context.Input.Duration
+
+Invoke-ActivityFunction -FunctionName "RequestApproval"
```
AnatoliB, line 11: This is supposed to be (eventually) working code. As opposed to samples in the docs that don't have to be directly runnable, this code should work as is, so we need to add at least very simple implementations of RequestApproval, ProcessApproval, and EscalateApproval functions. Very simple implementations that just log a message will do.

AnatoliB (approved):

PR #518, opened 2020-08-11
Added DurableTask abstract class and DurableTaskHandler

michaelpeng36:
Added DurableTask abstract class in preparation for awaiting DurableTimerTasks and ActivityInvocationTasks; the abstract class is extended by DurableTimerTask and ActivityInvocationTask.
Replaced DurableTimer and ActivityInvocationTracker classes with DurableTaskHandler, which operates on DurableTasks. Refactored pre-existing cmdlets and unit tests for respective tasks.
Added DurableTaskHandler unit tests (verifying proper functionality for the WaitAll function).


src/Durable/ActivityInvocationTask.cs
```diff
@@ -7,13 +7,63 @@
...
+    using WebJobs.Script.Grpc.Messages;
+
+    public class ActivityInvocationTask : DurableTask
     {
-        public string Name { get; }
+        internal string FunctionName { get; }
+
+        private object FunctionInput { get; }
+
+        public ActivityInvocationTask(string functionName, object functionInput)
```
AnatoliB: Should we make this constructor internal?

src/Durable/DurableTimerTask.cs
```diff
@@ -0,0 +1,65 @@
...
+    // Returned by the Start-DurableTimer cmdlet if the NoWait flag is present, representing a timeout task
+    // All DurableTimerTasks must be complete or canceled for the orchestration to complete
+    public class DurableTimerTask : DurableTask
+    {
+        private bool IsCompleted { get; }
+        private bool IsCanceled { get; set; }
+        public DateTime FireAt { get; set; }
+
+        // Only incomplete, uncanceled DurableTimerTasks should be created
+        public DurableTimerTask(
```
AnatoliB: Make this constructor internal as well?

src/Durable/DurableTimerTask.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    using System;
+    using System.Linq;
+
+    // Returned by the Start-DurableTimer cmdlet if the NoWait flag is present, representing a timeout task
+    // All DurableTimerTasks must be complete or canceled for the orchestration to complete
+    public class DurableTimerTask : DurableTask
+    {
+        private bool IsCompleted { get; }
+        private bool IsCanceled { get; set; }
```
AnatoliB: I can see this property assigned but not used. Do we need it?

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psd1
```diff
@@ -61,7 +61,7 @@ CmdletsToExport = @(
     'Set-FunctionInvocationContext',
     'Start-DurableTimer',
     'Trace-PipelineObject',
-    'Wait-ActivityFunction')
+    'Wait-DurableTask')
```
AnatoliB, line 64: Let's create an alias mapping the old name to the new one, so that we don't break existing code. We'll need something like `Set-Alias -Name Wait-ActivityFunction -Value Wait-DurableTask` in the .psm1, and add the alias name to the AliasesToExport entry below.

src/Durable/DurableTask.cs
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#pragma warning disable 1591 // "Missing XML comments for publicly visible type"
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System.Collections.Generic;
+    public abstract class DurableTask
```
AnatoliB, line 12: Please insert a blank line before the class definition.

src/Durable/DurableTaskHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using Utility;
+    internal class DurableTaskHandler
```
AnatoliB, line 13: A blank line before this one.

AnatoliB (changes requested):
A few minor comments, almost there.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psd1
michaelpeng36: Sounds good. I will then revert the changes to the TestFunctionApp orchestrators and the sample code where I replaced Wait-ActivityFunction with Wait-DurableTask.

examples/durable/DurableApp/FanOutFanInOrchestrator/run.ps1
```diff
@@ -9,7 +9,7 @@ $parallelTasks =
         Invoke-ActivityFunction -FunctionName 'SayHello' -Input $Name -NoWait
     }
 
-$output = Wait-DurableTask -Task $parallelTasks
+$output = Wait-ActivityFunction -Task $parallelTasks
```
AnatoliB: I think we should actually keep `Wait-DurableTask` in our examples and tests. When I asked for an alias, I was thinking about _external users_ code. Public Preview has been announced today, so external users will start writing code using `Wait-ActivityFunction` and we don't want to break it suddenly without a strong reason and an appropriate announcement. However, we do want to rename the command, so the code we directly control (examples and tests) should reflect this intention.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psd1
AnatoliB: See the comment at https://github.com/Azure/azure-functions-powershell-worker/pull/518#discussion_r473213770

src/Durable/ActivityInvocationTask.cs
AnatoliB: It is still public, are you making it internal?

AnatoliB (approved):

Issue #521, opened 2020-08-13
Change app directory on environment reload request


AnatoliB:
Some PowerShell apps intermittently fail on Azure complaining about not being able to app content under `D:\local\Temp\functions\standby`, typically when trying to load modules. The impact seems to be limited to apps that:
- include custom modules into the app content (_not_ using Managed Dependencies); **and**
- run from package.

This seems to be caused by the PowerShell worker _not_ changing the app directory when handling the environment reload request: https://github.com/Azure/azure-functions-powershell-worker/blob/ea4aaf77f4e2e5b64ba3cb1b81f29721f933e97c/src/RequestProcessor.cs#L345

The worker must change the current directory to the one specified by the environment reload request.

Until this is fixed, the workaround is to set the `WEBSITE_USE_PLACEHOLDER` app setting to `0`.

PR #522, opened 2020-08-14
Change current directory on environment reload request

AnatoliB:
Resolves issue #521

### DIFF
--- a/src/RequestProcessor.cs
+++ b/src/RequestProcessor.cs
@@ -357,9 +357,12 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
             var rpcLogger = new RpcLogger(_msgStream);
             rpcLogger.SetContext(request.RequestId, null);
 
-            var setCurrentDirMessage = string.Format(PowerShellWorkerStrings.SettingCurrentDirectory, environmentReloadRequest.FunctionAppDirectory);
-            rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, setCurrentDirMessage);
-            Directory.SetCurrentDirectory(environmentReloadRequest.FunctionAppDirectory);
+            if (environmentReloadRequest.FunctionAppDirectory != null)
+            {
+                var setCurrentDirMessage = string.Format(PowerShellWorkerStrings.SettingCurrentDirectory, environmentReloadRequest.FunctionAppDirectory);
+                rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, setCurrentDirMessage);
+                Directory.SetCurrentDirectory(environmentReloadRequest.FunctionAppDirectory);
+            }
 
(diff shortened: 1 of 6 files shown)


src/RequestProcessor.cs
```diff
@@ -355,6 +356,11 @@ internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessa
 
             var rpcLogger = new RpcLogger(_msgStream);
             rpcLogger.SetContext(request.RequestId, null);
+
+            var setCurrentDirMessage = string.Format(PowerShellWorkerStrings.SettingCurrentDirectory, environmentReloadRequest.FunctionAppDirectory);
+            rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, setCurrentDirMessage);
+            Directory.SetCurrentDirectory(environmentReloadRequest.FunctionAppDirectory);
```
Francisco-Gamino: Should we check that the directory exists before setting it? I remember that in the past for proxy call the working directory could be null.

Francisco-Gamino:
One minor comment. 
Also, I don't remember if for updating protobuff files the we need a separate PR just for that, so please follow-up with @pragnagopa. Other than that, LGTM.

Francisco-Gamino (approved):
One minor comment.
Also, I don't remember if for updating protobuff files the we need a separate PR just for that, so please follow-up with @pragnagopa. Other than that, LGTM.

src/RequestProcessor.cs
AnatoliB: Checking for null: yes, thank you, done.
Checking that the directory exists: I don't know if we have a better way of handling the situation when it does not exist. Perhaps the best we can do is trust the host and fail if it does not exist - this is what NodeJs and Python workers also do.

src/RequestProcessor.cs
```diff
@@ -355,6 +356,14 @@ internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessa
 
             var rpcLogger = new RpcLogger(_msgStream);
             rpcLogger.SetContext(request.RequestId, null);
+
+            if (environmentReloadRequest.FunctionAppDirectory != null)
+            {
```
pragnagopa: Please add unit test. Also, open issue here: https://github.com/Azure/azure-functions-integration-tests/issues/ - to track E2E test that verifies app directory after specialization.

pragnagopa (approved):
Thanks. LGTM. Added a minor comment on tests.

src/RequestProcessor.cs
pragnagopa: Yes. Trusting the host is OK. Any error handling should be done in  the host for this.

pragnagopa (approved):

src/RequestProcessor.cs
AnatoliB: Thank you. Added: https://github.com/Azure/azure-functions-integration-tests/issues/10

PR #527, opened 2020-08-21
Added support for external events and the ability to cancel an awaited durable timer

michaelpeng36:
Added Start-DurableExternalEventListener cmdlet, which listens for an external event to be raised indefinitely. If the NoWait flag is present, then it returns an external event task to be awaited with the Wait-DurableTask cmdlet.

Added Send-DurableExternalEvent cmdlet, which makes an HTTP POST request to send an external event (along with associated event data) to a given orchestration instance.

Added Stop-DurableTimerTask cmdlet. which marks a given durable timer task as canceled. Without this cmdlet, an orchestration will not complete until all timers created have expired.

Added unit tests for external events and an E2E test for all three cmdlets.


src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
```diff
@@ -141,3 +141,91 @@ function New-OrchestrationCheckStatusResponse {
...
+        [Parameter(
+            Position=4,
+            ValueFromPipelineByPropertyName=$true)]
+        [string] $ConnectionName
+    )
+
+    CheckIfDurableFunctionsEnabled
+    
+    $DurableClient = GetDurableClientFromModulePrivateData
+    if ($null -eq $DurableClient) {
```
AnatoliB: Let's move this check to the GetDurableClientFromModulePrivateData implementation, so we don't keep duplicating it.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psd1
```diff
@@ -59,7 +61,9 @@ CmdletsToExport = @(
     'Invoke-ActivityFunction',
     'Push-OutputBinding',
     'Set-FunctionInvocationContext',
+    "Start-DurableExternalEventListener",
```
AnatoliB: Single quotes, just for consistency with the rest

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
```diff
@@ -141,3 +141,91 @@ function New-OrchestrationCheckStatusResponse {
...
+            ValueFromPipelineByPropertyName=$true)]
+        [string] $TaskHubName,
+
+        [Parameter(
+            Position=4,
+            ValueFromPipelineByPropertyName=$true)]
+        [string] $ConnectionName
+    )
+
+    CheckIfDurableFunctionsEnabled
```
AnatoliB: I've just removed this function, no need to call it anymore

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
```diff
@@ -141,3 +141,91 @@ function New-OrchestrationCheckStatusResponse {
...
+            ValueFromPipelineByPropertyName=$true)]
+        [string] $EventName,
+
+        [Parameter(
+            Position=2,
+            ValueFromPipelineByPropertyName=$true)]
+        [object] $EventData,
+
+		[Parameter(
+            Position=3,
```
AnatoliB: It is usually a good idea to be careful with positional parameters, assign positions to the most obvious and intuitive parameters only, and limit the number of positional parameters to 1-3. Otherwise, this encourages writing code passing a bunch of unnamed parameters, which becomes error-prone and hard to read. In this specific case, I recommend removing positions from TaskHubName and ConnectionName, so the user will have to specify parameter names explicitly.

test/Unit/Durable/DurableTaskHandlerTests.cs
```diff
@@ -161,7 +161,7 @@ public void WaitAny_OutputsEarliestCompletedTask_WhenAnyTaskCompleted(bool compl
         }
 
         [Theory]
-        [InlineData(false, true)]
+        // [InlineData(false, true)]
```
AnatoliB: Why is it disabled?

AnatoliB (changes requested):
Great job Michael, thank you. Please let me know if you want to address the feedback soon, or I will do it next week.

AnatoliB (approved):

PR #532, opened 2020-08-31
Fix flaky E2E durable tests

AnatoliB:
Return the file log content from the function instead of the file name


Francisco-Gamino (approved):
LGTM

PR #535, opened 2020-09-03
Warm up PowerShell Worker before receiving function load request

### LINKED ISSUE #485: Warm up PowerShell Worker before receiving function load request
When the worker starts on an Antares placeholder, it spends some time just waiting for requests, as it does not belong to any app yet. It could use this time to perform some app-independent warm-up: for example, create a PS runspace, load some app-independent modules and assemblies, open and read some files to make sure they are in the cache, etc. When the placeholder is assigned to an app and the real work starts, this preliminary warm-up should reduce the cold start time.

AnatoliB:
Resolves #485

- The first PowerShell instance creation is moved to `Worker.Main`, and it is now performed immediately after starting the process, not waiting for the function load request. It is important to remember that at this point we cannot initialize `$env:PSModulePath` because the app root path is not resolved yet. So, for the first PowerShell instance, this step is deferred until the function load request and performed in `RequestProcessor.SetupAppRootPathAndModulePath`. For all the subsequent PowerShell instances, `$env:PSModulePath` comes from the initial session state, as before.
- Handling the function load request involves creating a PowerShell function from a script file (see `PowerShellManager.DeployAzFunctionToRunspace`), and this takes hundreds of milliseconds even for a short script when performed for the first time. However, the subsequent invocations are significantly faster. So, creating and removing a dummy function proactively saves noticeable time on the function load request (see `Worker.WarmUpPowerShell`).

This reduces cold start time by about 4 seconds (measured on App Service plan, S1 instance):

Stage | Before the change (ms) | After the change (ms)
--- | --- | ---
Function load request | 3700 | 200
First function invocation request | 2200 | 1700
**Total** | **5900** | **1900**

### DIFF
--- a/src/RequestProcessor.cs
+++ b/src/RequestProcessor.cs
@@ -199,8 +199,6 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
                     // Setup the FunctionApp root path and module path.
                     FunctionLoader.SetupWellKnownPaths(functionLoadRequest, managedDependenciesPath);
 
-                    LogPowerShellVersion(rpcLogger, _firstPwshInstance);
-
                     _firstPwshInstance.AddCommand("Set-Content")
                         .AddParameter("Path", "env:PSModulePath")
                         .AddParameter("Value", FunctionLoader.FunctionModulePath)
@@ -495,12 +493,6 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
             {
                 response.ReturnValue = results[AzFunctionInfo.DollarReturn].ToTypedData();
             }
-        }
-
-        private static void LogPowerShellVersion(RpcLogger rpcLogger, System.Management.Automation.PowerShell pwsh)
-        {
-            var message = string.Format(PowerShellWorkerStrings.PowerShellVersion, Utils.GetPowerShellVersion(pwsh));
-            rpcLogger.Log(isUserOnlyLog: false, LogLevel.Information, message);
         }
 
         #endregion
--- a/src/RequestProcessor.cs
+++ b/src/RequestProcessor.cs
@@ -497,7 +497,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
                 throw new InvalidOperationException(PowerShellWorkerStrings.FunctionAppRootNotResolved);
             }
 
-            _firstPwshInstance.AddCommand("Set-Content")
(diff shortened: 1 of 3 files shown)


src/RequestProcessor.cs
```diff
@@ -493,10 +488,19 @@ private static void BindOutputFromResult(InvocationResponse response, AzFunction
...
-            var message = string.Format(PowerShellWorkerStrings.PowerShellVersion, Utils.GetPowerShellVersion(pwsh));
-            rpcLogger.Log(isUserOnlyLog: false, LogLevel.Information, message);
+            FunctionLoader.SetupWellKnownPaths(functionLoadRequest, managedDependenciesPath);
+
+            if (FunctionLoader.FunctionAppRootPath == null)
+            {
+                throw new InvalidOperationException(PowerShellWorkerStrings.FunctionAppRootNotResolved);
+            }
+
+            _firstPwshInstance.AddCommand("Set-Content")
```
Francisco-Gamino: Please include the module name with the cmdlet, e.g., `"Microsoft.PowerShell.Management\\Set-Content"`

src/Worker.cs
```diff
@@ -34,18 +36,50 @@ public async static Task Main(string[] args)
...
         }
+
+        // Warm up the PowerShell instance so that the subsequent function load and invocation requests are faster
+        private static void WarmUpPowerShell(System.Management.Automation.PowerShell firstPowerShellInstance)
+        {
+            // It turns out that creating/removing a function warms up the runspace enough.
+            // We just need this name to be unique, so that it does not coincide with an actual function.
+            const string DummyFunctionName = "DummyFunction-71b09c92-6bce-42d0-aba1-7b985b8c3563";
+
+            firstPowerShellInstance.AddCommand("New-Item")
```
Francisco-Gamino: Same for `Microsoft.PowerShell.Management\\New-Item` and `Microsoft.PowerShell.Management\\Remove-Item`

Francisco-Gamino (approved):
Thanks @AnatoliB for sending this out!

A couple of minor comments. Other than that, LGTM.

src/RequestProcessor.cs
AnatoliB: Done.

src/Worker.cs
AnatoliB: Done.

PR #530, opened 2020-08-26
Add conditional stage to upload the generated nuget package to the UploadPackageToPreReleaseFeed.

Francisco-Gamino:
This PR contains the following changes:
* Added logic to evaluate the AzureFunctionsPreReleaseFeed variable and decide whether the generated nuget package will be uploaded to the AzureFunctionsPreReleaseFeed.
* Added conditional stage to upload the generated nuget package to the UploadPackageToPreReleaseFeed.

PR #556, opened 2020-11-04
Add conditional stage to upload the generated nuget package to the UploadPackageToPreReleaseFeed. (#530)

Francisco-Gamino:
* Evaluate the AzureFunctionsPreReleaseFeed variable and decide whether the generated nuget package will be uploaded to the AzureFunctionsPreReleaseFeed

* Add conditional stage to upload the PowerShell worker nuget package to the AzureFunctionsPreReleaseFeed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -59,8 +59,18 @@ steps:
       Copy-Item -Path $sourcePath -Destination $(Build.ArtifactStagingDirectory) -ErrorAction Stop -Verbose -Force
   displayName: 'Copy package to ArtifactStagingDirectory'
 
+- pwsh: |
+    $uploadPackage = $null
+    if (-not ([bool]::TryParse($env:UPLOADPACKAGETOPRERELEASEFEED, [ref] $uploadPackage)))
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -59,12 +59,31 @@ steps:
       Copy-Item -Path $sourcePath -Destination $(Build.ArtifactStagingDirectory) -ErrorAction Stop -Verbose -Force
   displayName: 'Copy package to ArtifactStagingDirectory'
 
+- pwsh: |
```
AnatoliB, line 62: Do we already have (and do we need?) the same code on the `dev` branch already?

AnatoliB (approved):

azure-pipelines.yml
Francisco-Gamino: Yes, this code already exists in the dev branch -- this is how we build PowerShell 7 for V3. Without this porting this change to `v3.x/ps6`, we cannot push the package to the `AzureFunctionPreReleaseFeed`.

PR #558, opened 2020-11-13
Improve Managed Dependencies logs

AnatoliB:
Improving Managed Dependencies logs:
- always log installed snapshot content, regardless of whether this snapshot is to be promoted or not, so that we can count all the modules retrieved from the PS Gallery;
- log the current snapshot content periodically (once a day), so that it is readily available even if there are no new snapshots installed or promoted.


Francisco-Gamino (approved):
LGTM

Issue #559, opened 2020-11-14
Using Invoke-AzResourceAction causes Trace-PipelineObject to throw an error


janegilring:
We are using [Invoke-AzResourceAction](https://docs.microsoft.com/en-us/powershell/module/az.resources/invoke-azresourceaction?view=azps-5.0.0) in an Azure Function. The command succeeds, but always returns an error - even inside a try/catch. Works without any errors locally - hence I suppose the issue is related to the Azure Functions PowerShell Worker.

The error is:
> ERROR: Trace-PipelineObject : Cannot bind argument to parameter 'InputObject' because it is null. + CategoryInfo : InvalidData: (:) [Trace-PipelineObject], ParameterBindingValidationException + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Microsoft.Azure.Functions.PowerShellWorker.Commands.TracePipelineObjectCommand Script stack trace: at <ScriptBlock>, D:\home\site\wwwroot\BackupFileShare\run.ps1: line 62 System.Management.Automation.ParameterBindingValidationException: Cannot bind argument to parameter 'InputObject' because it is null. at System.Management.Automation.CmdletParameterBinderController.BindValueFromPipeline(PSObject inputToOperateOn, MergedCompiledCommandParameter parameter, ParameterBindingFlags flags) at System.Management.Automation.CmdletParameterBinderController.BindUnboundParametersForBindingStateInParameterSet(PSObject inputToOperateOn, CurrentlyBinding currentlyBinding, UInt32 validParameterSets) at System.Management.Automation.CmdletParameterBinderController.BindUnboundParametersForBindingState(PSObject inputToOperateOn, CurrentlyBinding currentlyBinding, UInt32 validParameterSets) at System.Management.Automation.CmdletParameterBinderController.BindPipelineParametersPrivate(PSObject inputToOperateOn) at System.Management.Automation.CmdletParameterBinderController.BindPipelineParameters(PSObject inputToOperateOn) at System.Management.Automation.CommandProcessor.Read()

AnatoliB:
@janegilring Could you please share a code sample? Please try to either capture the output of Invoke-AzResourceAction into a variable, or pipe it to Out-Null, and let us know if this helps as a workaround.

janegilring:
@AnatoliB Here is a code sample:

```powershell

$null = Set-AzContext $AzSubscriptionId

[...]

Issue #504, opened 2020-07-30
Orchestrator Should Not Throw an Exception if Output is $null


michaelpeng36:
In the case that the orchestrator returns a $null object, for example as in the following orchestrator:
```
using namespace System.Net

param($Context)

$output
```

it throws the following error

```[7/30/2020 5:38:19 PM] ERROR: Cannot bind argument to parameter 'InputObject' because it is null.
[7/30/2020 5:38:19 PM]
[7/30/2020 5:38:19 PM] Exception             : 
[7/30/2020 5:38:19 PM]     Type              : System.Management.Automation.ParameterBindingValidationException
[7/30/2020 5:38:19 PM]     Message           : Cannot bind argument to parameter 'InputObject' because it is null.
[7/30/2020 5:38:19 PM]     ParameterName     : InputObject
[7/30/2020 5:38:19 PM]     ParameterType     : System.Object
[7/30/2020 5:38:19 PM]     ErrorId           : ParameterArgumentValidationErrorNullNotAllowed
[7/30/2020 5:38:19 PM]     CommandInvocation : 
[7/30/2020 5:38:19 PM]         MyCommand        : Trace-PipelineObject
[7/30/2020 5:38:19 PM]         HistoryId        : 1
[7/30/2020 5:38:19 PM]         InvocationName   : Microsoft.Azure.Functions.PowerShellWorker\Trace-PipelineObject
[7/30/2020 5:38:19 PM]         PipelineLength   : 2
[7/30/2020 5:38:19 PM]         PipelinePosition : 2
[7/30/2020 5:38:19 PM]         ExpectingInput   : True
[7/30/2020 5:38:19 PM]     ErrorRecord       : 
[7/30/2020 5:38:19 PM]         Exception             : 
[7/30/2020 5:38:19 PM]             Type    : System.Management.Automation.ParentContainsErrorRecordException
[7/30/2020 5:38:19 PM]             Message : Cannot bind argument to parameter 'InputObject' because it is null.
[7/30/2020 5:38:19 PM]             HResult : -2146233087
[7/30/2020 5:38:19 PM]         CategoryInfo          : InvalidData: (:) [Trace-PipelineObject], ParentContainsErrorRecordException
[7/30/2020 5:38:19 PM]         FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Microsoft.Azure.Functions.PowerShellWorker.Commands.TracePipelineObjectCommand
[...]

PR #566, opened 2020-11-30
Allow writing null to Function output

### LINKED ISSUE #561: Writing $null to the output causes Trace-PipelineObject error
When any Function writes `$null` to the output stream, this causes the following error:
```
ERROR: Trace-PipelineObject : Cannot bind argument to parameter 'InputObject' because it is null.
```

The `$null` value can be hardcoded, returned from a cmdlet or a function, etc. For example, each of the following lines causes this error:

``` PowerShell
$null
Write-Output $null
Get-Something  # This command returns $null
```

## Workaround

Avoid writing `$null` to output. If you invoke a command that may return `$null`, capture the value by assigning it to a variable, or pipe it to Out-Null, for example:

``` PowerShell
$Result = Do-Something
Get-Something | Out-Null
```

AnatoliB:
Allow writing null to Function output by making the InputObject parameter of Trace-PipelineObject optional (which enables passing $null).

Resolves #561, #559, #504

### DIFF
--- a/src/Public/Commands/TracePipelineObjectCommand.cs
+++ b/src/Public/Commands/TracePipelineObjectCommand.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Commands
         /// <summary>
         /// The object from pipeline.
         /// </summary>
-        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true)]
+        [Parameter(Position = 0, ValueFromPipeline = true)]
         public object InputObject { get; set; }
 
         private static PowerShell s_pwsh;


Francisco-Gamino (approved):
LGTM

VasylZvarydchuk:
That's a great fix @AnatoliB. Thank you very much. We spent tons of time to refactor our PowerShell functions previously worked fine in Azure Automation. It took much time to investigate these weird errors and add `Out-Null` statements. How to know when this fix will be release?

AnatoliB:
@VasylZvarydchuk Expect this to be deployed to Azure within ~3-4 weeks.

VasylZvarydchuk:
@AnatoliB we still see such errors. Apparently, this change doesn't fix the problem. So, if code has such code like `return $null` PowerShell Azure Functions fails with `Cannot bind argument to parameter 'InputObject' because it is null`

AnatoliB:
@VasylZvarydchuk The 3-4 weeks estimate turned out to be overoptimistic. For reasons unrelated to the fix, the rollout to Azure was delayed. It is happening right now, and many regions already have the fix. Please check the Functions runtime version for your app on the Portal (the Overview tab): if you see 3.0.15371 or higher, you've got the fix. Otherwise, you may need to wait for a few more days and/or restart the app.

If you are on 3.0.15371 and still see the problem, please file a new issue, as this is probably something else.

VasylZvarydchuk:
Thank you @AnatoliB . I see we have still 3.0.15193.0. Ok, waiting.

VasylZvarydchuk:
@AnatoliB just today we have got a new AF update and indeed the fix works. Thank you very much

PR #571, opened 2020-12-19
Propagate durable activity failures to orchestrator function

### LINKED ISSUE #572: Durable Function activity failures are difficult to handle
When an activity function fails because of an exception, this exception is not propagated to the orchestrator function. For example, the `Invoke-ActivityFunction` call below will not throw any exception, will not write anything to the Error stream, the orchestration function will continue execution, and the orchestration will reach the 'Completed' status.

**Activity function (MyActivity):**
``` PowerShell
throw 'Something is wrong'
```

**Orchestrator function:**
``` PowerShell
Invoke-ActivityFunction -FunctionName 'MyActivity'
...
```
As a result, there is no clean and simple way to handle failures. It is technically possible to check the activity function logs (as the exception will be reported there), or use other means to check for other side effects to understand whether the activity succeeded or not. However, these solutions are complicated and should be invented on the case-by-case basis. Instead, we should make `Invoke-ActivityFunction` report regular PowerShell errors and let the orchestrator function use regular PowerShell language constructs to handle these errors.

AnatoliB:
Resolves #572 

There are two problems to solve here:

1. Make `Invoke-ActivityFunction` report an error on activity failure. This part is simple: `TaskFailed` events are recorded in the orchestration history, so we just make `Invoke-ActivityFunction` write an error (`ActivityFailureException`) to the Error stream. By default, this is treated as a non-terminating error in PowerShell, but the orchestrator function author can use either `$ErrorActionPreference = 'Stop'` or `-ErrorAction Stop` to convert it to an exception, if desired.

2. Unfortunately, completing (1) is not enough: throwing this exception from the orchestrator function leads to errors complaining about non-deterministic orchestrator behavior because of certain implementation quirks affecting out-of-proc language workers. This PR follows the approach taken by the JavaScript Durable implementation (https://github.com/Azure/azure-functions-durable-js/pull/145): the orchestration replay state is embedded into the exception message after the `$OutOfProcData$` marker.

### DIFF
--- a/src/Durable/OrchestrationFailureException.cs
+++ b/src/Durable/OrchestrationFailureException.cs
@@ -9,6 +9,11 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
     using System.Collections.Generic;
     using Newtonsoft.Json;
 
+    /// <summary>
+    /// OrchestrationFailureException should be propagated back to the Host when an orchestrator function
+    /// throws and does not handle an exception. The Durable Functions extension implementation requires
+    /// this exception message to contain the Json-serialized orchestration replay state after a special marker.
+    /// </summary>
     internal class OrchestrationFailureException : Exception
     {
         public const string OutOfProcDataLabel = "\n\n$OutOfProcData$:";
@@ -24,6 +29,9 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
 
         private static string FormatOrchestrationFailureMessage(List<OrchestrationAction> actions, Exception exception)
         {
+            // For more details on why this message looks like this, see:
+            // - https://github.com/Azure/azure-functions-durable-js/pull/145
+            // - https://github.com/Azure/azure-functions-durable-extension/pull/1171
             var orchestrationMessage = new OrchestrationMessage(isDone: false, new List<List<OrchestrationAction>> { actions }, output: null, exception.Message);
             var message = $"{exception.Message}{OutOfProcDataLabel}{JsonConvert.SerializeObject(orchestrationMessage)}";
             return message;
--- a/src/Durable/OrchestrationFailureException.cs
+++ b/src/Durable/OrchestrationFailureException.cs
@@ -1,0 +1,32 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    internal class OrchestrationFailureException : Exception
+    {
+        public const string OutOfProcDataLabel = "\n\n$OutOfProcData$:";
+
+        public OrchestrationFailureException()
+        {
+        }
+
+        public OrchestrationFailureException(List<OrchestrationAction> actions, Exception innerException)
+            : base(FormatOrchestrationFailureMessage(actions, innerException), innerException)
+        {
+        }
+
+        private static string FormatOrchestrationFailureMessage(List<OrchestrationAction> actions, Exception exception)
+        {
+            var orchestrationMessage = new OrchestrationMessage(isDone: false, new List<List<OrchestrationAction>> { actions }, output: null, exception.Message);
+            var message = $"{exception.Message}{OutOfProcDataLabel}{JsonConvert.SerializeObject(orchestrationMessage)}";
+            return message;
+        }
+    }
+}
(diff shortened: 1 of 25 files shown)


src/Durable/OrchestrationFailureException.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+        }
+
+        private static string FormatOrchestrationFailureMessage(List<OrchestrationAction> actions, Exception exception)
+        {
+            // For more details on why this message looks like this, see:
+            // - https://github.com/Azure/azure-functions-durable-js/pull/145
+            // - https://github.com/Azure/azure-functions-durable-extension/pull/1171
+            var orchestrationMessage = new OrchestrationMessage(isDone: false, new List<List<OrchestrationAction>> { actions }, output: null, exception.Message);
+            var message = $"{exception.Message}{OutOfProcDataLabel}{JsonConvert.SerializeObject(orchestrationMessage)}";
```
Francisco-Gamino, line 36: Is value `orchestrationMessage` always a valid JSON?

test/E2E/Azure.Functions.PowerShellWorker.E2E/Azure.Functions.PowerShellWorker.E2E/DurableEndToEndTests.cs
```diff
@@ -87,6 +87,55 @@ public async Task DurableClientFollowsAsyncPattern()
...
+            using var httpClient = new HttpClient();
+
+            while (true)
+            {
+                var statusResponse = await httpClient.GetAsync(statusQueryGetUri);
+                switch (statusResponse.StatusCode)
+                {
+                    case HttpStatusCode.Accepted:
+                    {
+                            if (DateTime.UtcNow > startTime + orchestrationCompletionTimeout)
```
Francisco-Gamino: Nit: Extra tab before the `if` statement.

test/Unit/Durable/OrchestrationInvokerTests.cs
```diff
@@ -122,6 +122,18 @@ public void ReturnsNoInvocationOutputWhenStopped()
...
+        {
+            Exception originalException = new Exception("Original exception");
+            _mockPowerShellServices.Setup(_ => _.EndInvoke(It.IsAny<IAsyncResult>())).Throws(originalException);
+
+            var output = new[] { "item1", "item2" };
+
+            var thrownException = Assert.Throws<OrchestrationFailureException>(() => InvokeOrchestration(completed: true, output));
+            Assert.Same(originalException, thrownException.InnerException);
+        }
+
```
Francisco-Gamino, line 136: Should we have a test case that validates that the `FullyQualifiedErrorId` contains `Functions.Durable.ActivityFailure`? I can see this becoming a popular search online later on.

Francisco-Gamino (approved):
A couple of minor comments. Other than that, LGTM.

src/Durable/OrchestrationFailureException.cs
AnatoliB: Please note we are converting _to_ JSON, not _from_ JSON. `orchestrationMessage` is an object, and the result of this conversion is expected to be a valid JSON (assuming Newtonsoft.Json is correct).

test/Unit/Durable/OrchestrationInvokerTests.cs
AnatoliB: Good point. We cannot validate this right here because `FullyQualifiedErrorId` does not belong to the exception. It's the other way around: the exception is wrapped into an `ErrorRecord`, and this `ErrorRecord` contains `FullyQualifiedErrorId`... But I agree we should emphasize this because this errorId will become a part of the public contract that we shouldn't break in the future. I think this check should go to `DurableActivityErrorHandler` unit tests, I'll do that.

test/E2E/Azure.Functions.PowerShellWorker.E2E/Azure.Functions.PowerShellWorker.E2E/DurableEndToEndTests.cs
AnatoliB: Thanks, fixed

src/Durable/OrchestrationFailureException.cs
Francisco-Gamino: You are correct Anatoli, I missed read it. My comment does not apply. Resolving.

test/Unit/Durable/OrchestrationInvokerTests.cs
Francisco-Gamino: Sounds good. Thanks Anatoli.

PR #573, opened 2020-12-21
Update pipeline trigger to only include `releases/*` branches

Francisco-Gamino:
Update pipeline trigger to only include `releases/*` branches.


AnatoliB (approved):
LGTM

PR #576, opened 2020-12-23
Fix orchestration termination

### LINKED ISSUE #575: Orchestrations don't terminate properly
When a Durable orchestration is terminated by an external signal (e. g. using the [termination URI](https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-instance-management?tabs=csharp#retrieve-http-management-webhook-urls), the orchestration runtime status correctly changes to "Terminated", but the orchestrator function continues running, potentially invoking more activities.

For example, consider the following orchestrator code:

 ``` PowerShell
Invoke-ActivityFunction -Name 'A'
Write-Warning 'Still running'
Invoke-ActivityFunction -Name 'B'
Write-Warning 'Yes, still running'
Invoke-ActivityFunction -Name 'C'
```

If the termination signal arrives while activity A is running:

- **Expected behavior**: activities B and C do not run, the "still running" warnings are not emitted.
- **Actual behavior**: the orchestrator runs to the end, scheduling activities B and C, emitting both warnings, but not actually waiting for any activity to finish.

This does not repro with Durable Functions in C# or JavaScript, so this is specific to the PowerShell implementation.

AnatoliB:
Resolves #575 

When the durable extension receives a termination signal, it invokes the orchestrator, which replays execution until the activity that is still running. In the execution history, this activity is marked as _scheduled_ but not _completed._ As a result, we need to stop the orchestrator, in exactly the same way as when the activity has not been scheduled yet.

### DIFF
--- a/src/Durable/DurableTaskHandler.cs
+++ b/src/Durable/DurableTaskHandler.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
                             break;
                     }
                 }
-                else if (scheduledHistoryEvent == null)
+                else
                 {
                     InitiateAndWaitForStop(context);
                 }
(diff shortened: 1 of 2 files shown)


Francisco-Gamino (approved):
LGTM

PR #580, opened 2021-01-06
Move files to Durable subfolders

AnatoliB:
Just moving files (and updating namespaces accordingly) to give the Durable implementation code more structure.


src/Durable/Commands/InvokeActivityFunctionCommand.cs
```diff
@@ -1,55 +1,56 @@
...
+    /// Invoke an activity function.
+    /// </summary>
+    [Cmdlet("Invoke", "ActivityFunction")]
+    public class InvokeActivityFunctionCommand : PSCmdlet
+    {
+        /// <summary>
+        /// Gets and sets the activity function name.
+        /// </summary>
+        [Parameter(Mandatory = true)]
+        public string FunctionName { get; set; }
```
Francisco-Gamino, line 24: Should we validate `FunctionName` for null or empty?

AnatoliB: I believe `Mandatory = true` is already enforcing the check. Furthermore, we later verify that this name actually matches an existing function name.

Francisco-Gamino (approved):
One minor comment. Otherwise, LGTM.

src/Durable/Commands/InvokeActivityFunctionCommand.cs
AnatoliB: Please also note that this PR shows that the entire file changed just because of line ending unification. The actual meaningful changes are only around namespaces and file locations.

src/Durable/Commands/SetFunctionInvocationContextCommand.cs
```diff
@@ -1,57 +1,57 @@
...
-                    {
-                        privateData.Remove(ContextKey);
-                        privateData.Remove(DurableClientKey);
-                    }
-                    break;
-            }
-        }
-    }
-}
+//
```
AnatoliB, line 1: This file was also just moved to a subfolder, and then git unified line endings to match .gitattributes, this is why the PR shows that the entire file changed.

src/Durable/Commands/SetFunctionInvocationContextCommand.cs
```diff
@@ -1,57 +1,57 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#pragma warning disable 1591 // Missing XML comment for publicly visible type or member 'member'
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
```
AnatoliB: I forgot to change the namespace to *.Commands, will do.

PR #592, opened 2021-02-01
[Durable] Custom orchestration status

AnatoliB:
Feature documentation: [Custom orchestration status](https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-custom-orchestration-status).

Usage sample:
``` PowerShell
Set-DurableCustomStatus -CustomStatus 'Processing'
```

### DIFF
--- a/src/Durable/Commands/SetDurableCustomStatusCommand.cs
+++ b/src/Durable/Commands/SetDurableCustomStatusCommand.cs
@@ -1,0 +1,29 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#pragma warning disable 1591 // Missing XML comment for publicly visible type or member 'member'
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable.Commands
+{
+    using System.Collections;
+    using System.Management.Automation;
+
+    [Cmdlet("Set", "DurableCustomStatus")]
+    public class SetDurableCustomStatusCommand : PSCmdlet
+    {
+        [Parameter(
+            Mandatory = true,
+            Position = 0,
+            ValueFromPipeline = true)]
+        public object CustomStatus { get; set; }
+
+        protected override void EndProcessing()
+        {
+            var privateData = (Hashtable)MyInvocation.MyCommand.Module.PrivateData;
+            var context = (OrchestrationContext)privateData[SetFunctionInvocationContextCommand.ContextKey];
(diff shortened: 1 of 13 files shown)


src/Durable/Commands/SetDurableCustomStatusCommand.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    using System.Collections;
+    using System.Management.Automation;
+
+    [Cmdlet("Set", "DurableCustomStatus")]
+    public class SetDurableCustomStatusCommand : PSCmdlet
+    {
+        [Parameter(
+            Mandatory = true,
+            Position = 0,
```
Francisco-Gamino, line 18: Do we need to assign a `Position` for this parameter? I took a look at the other Durable cmdlets and this is the only one with a Positional parameter.

src/Durable/Commands/SetDurableCustomStatusCommand.cs
```diff
@@ -0,0 +1,29 @@
...
+            ValueFromPipeline = true)]
```
Francisco-Gamino, line 19: What would the use case be here for `ValueFromPipeline`? Something like: 

```powershell
# 1)
'Processing Tokyo' | Set-DurableCustomStatus

# 2)
 @{ ProgressMessage = 'Processing Seattle'; Stage = 2 }  | Set-DurableCustomStatus

```

Francisco-Gamino (approved):
One minor comment and one question. Other than that, LGTM.

src/Durable/Commands/SetDurableCustomStatusCommand.cs
AnatoliB: I'm trying to be very careful with adding positional parameters in general, but in this case I believe this is justified. The cmdlet takes just one parameter, and the purpose of this parameter is very obvious, so I want to allow users omit the parameter name: `Set-DurableCustomStatus "my status"`.

I think we can carefully allow _some_ positional parameters for _some_ other durable-related cmdlets, but we can do this later.

src/Durable/Commands/SetDurableCustomStatusCommand.cs
AnatoliB: Yes, exactly.

src/Durable/Commands/SetDurableCustomStatusCommand.cs
Francisco-Gamino: Sounds good @AnatoliB, thank you.

PR #603, opened 2021-02-27
Handle prerelease module version specifications

### LINKED ISSUE #602: When a prerelease module version is specified in requirements.psd1, every PowerShell worker process installs it's own dependencies snapshot
**Problem**

When a prerelease module version is specified in `requirements.psd1`, every PowerShell worker process installs it's own dependencies snapshot, even if one is installed already. This leads to very long cold start on each PowerShell worker unnecessarily.

For example, if `requirements.psd1` contains:

``` PowerShell
@{
    'Az.Billing' = '4.0.2-preview'
}
```

the worker will successfully install a dependency snapshot on the first function invocation, and all the subsequent invocations routed to this worker will start fast and use this module. However, other worker process instances will not be able to locate the same dependency snapshot and will create another one.

**UPDATE (3/12):** One more consequence of this bug is that old dependency snapshots may not get purged anymore in some configurations.

**Root cause**

Installing the module will create the following folder structure:

```
    ../ManagedDependencies
        Az.Billing
            4.0.2
                module content files
```

However, the code looking for an acceptable snapshot will look for a folder named `4.0.2-preview` under `Az.Billing`. Since it is not present, the worker will assume that this snapshot is not good enough and will install another one.

**Workarounds**

Until the bug is fixed:
- avoid using prerelease modules; or
- include module content into the app and remove modules from requrements.psd1; or
- periodically (once a week or so), purge managed dependencies manually: stop the app, remove the entire ManagedDependencies folder, then start the app.

AnatoliB:
Resolves #602

### DIFF
--- a/src/DependencyManagement/InstalledDependenciesLocator.cs
+++ b/src/DependencyManagement/InstalledDependenciesLocator.cs
@@ -38,7 +38,9 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
             {
                 case VersionSpecificationType.ExactVersion:
                     return _storage.IsModuleVersionInstalled(
-                        snapshotPath, dependency.Name, dependency.VersionSpecification);
+                                snapshotPath, dependency.Name, dependency.VersionSpecification)
+                        || _storage.IsModuleVersionInstalled(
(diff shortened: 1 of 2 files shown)


Francisco-Gamino (approved):
LGTM

PR #604, opened 2021-02-27
Log the number of dependency snapshots installed and how many snapshots to keep

### LINKED ISSUE #601: Managed Dependencies: when purging, log the current number of dependency snapshots installed and how many (snapshots) we should keep
When purging, log the current number of dependency snapshots installed and how many (snapshots) we should keep. The number of snapshots to keep is determined by the app setting `MDMinNumberOfSnapshotsToKeep` or 1 if the setting is not preset.

The log statement should be added at https://github.com/Azure/azure-functions-powershell-worker/blob/dev/src/DependencyManagement/DependencySnapshotPurger.cs#L75 right before we iterate through the list of snapshots. 

The log message would be something like this:
`Number of dependency snapshots installed: <number>. Dependency snapshots to keep: <number>`

Francisco-Gamino:
Resolves https://github.com/Azure/azure-functions-powershell-worker/issues/601

### DIFF
--- a/src/DependencyManagement/DependencySnapshotPurger.cs
+++ b/src/DependencyManagement/DependencySnapshotPurger.cs
@@ -73,6 +73,12 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
                                             .OrderBy(entry => entry.Item2)
                                             .ToArray();
 
+            var snapshotsLogmessage = string.Format(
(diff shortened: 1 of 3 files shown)


AnatoliB (approved):
LGTM

Francisco-Gamino:
This PR is replaced by https://github.com/Azure/azure-functions-powershell-worker/pull/605.

PR #619, opened 2021-04-16
Install both 3.1 and 5.0 versions of dotnet SDK

AnatoliB:
We need .NET SDK 5.0 for the updated C# compiler.
We need .NET SDK 3.1 for running the tests, as we still build against the 3.1 framework.
This PR makes `build.ps1` check if both are installed, and install if necessary. It also simplifies the checking logic.


Francisco-Gamino (approved):
LGTM

PR #623, opened 2021-04-20
Fix intermittent durable e2e test failures

AnatoliB:
Sometimes Functions run very slowly in the e2e testing environment, so increasing the timeouts


Francisco-Gamino (approved):
LGTM

PR #624, opened 2021-04-20
Suppress tracing output of Durable Activity functions

### LINKED ISSUE #622: Activity function output is always logged
When an Activity function writes to Output, this data becomes the return value of the Activity, by definition. This is different from how Output is treated in non-Activity functions: data written to Output is logged, but the function must explicitly invoke `Push-OutputBinding` for the function to actually propagate this data somewhere else.

The problem is that Activity function Output is not just returned to the invoker, but _it is also logged_, like we do for non-Activity functions. This may lead to unintentional information disclosure, and just pollute logs with unneeded data. There is no way to selectively turn off logging of Activity output data.

We've intentionally made Activity functions special, and Output is treated differently. So, we should go all the way and exclude this data from logs. If the user does want this data to be traced, they are free to do this explicitly with `Write-Verbose` etc.

AnatoliB:
Resolves #622

### DIFF
--- a/src/Durable/DurableController.cs
+++ b/src/Durable/DurableController.cs
@@ -109,6 +109,11 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
             return true;
         }
 
+        public bool ShouldSuppressPipelineTraces()
+        {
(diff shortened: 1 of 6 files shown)


Francisco-Gamino (approved):
LGTM

PR #625, opened 2021-04-21
Rename Durable commands

AnatoliB:
Renaming some Durable Functions commands to follow the following pattern consistently:

_[Verb]_-**Durable**_[Additional qualifiers]_

**Old name** | **New name**
--- | ---
Invoke-ActivityFunction | Invoke-DurableActivity
New-OrchestrationCheckStatusResponse | New-DurableOrchestrationCheckStatusResponse
Start-NewOrchestration | Start-DurableOrchestration

The old names are kept as aliases for backward-compatibility purposes.


Francisco-Gamino (approved):
Given that the examples use the new cmdlet names, could you please add a test to validate that the alias are properly map? This would be to make sure that we do not break customer in the  future.
Otherwise, LGTM.

AnatoliB:
> Given that the examples use the new cmdlet names, could you please add a test to validate that the alias are properly map? This would be to make sure that we do not break customer in the future.

Great idea, thank you, I'll do this.

PR #626, opened 2021-04-21
Use local RPC endpoint when starting an orchestration

### LINKED ISSUE #597: Add support for the local RPC endpoint
The Durable Extension recently introduced a [local rpc endpoint](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.extensions.durabletask.durabletaskoptions.localrpcendpointenabled?view=azure-dotnet) feature that enables Durable SDKs to communicate directly to the Durable Functions API, whereas before this communicate would occur via HTTP calls. Without this feature, we cannot integrate Durable Functions for PowerShell with EasyAuth, as framework-internal HTTP calls will fail due to authentication requirements.

**Implementation Requirements**
Thankfully, it appears that supporting this feature is relatively simple: it simply means that we call a different URI in places where we used to make HTTP requests. Also, it **appears** that this feature is only used for durable clients bindings, which potentially limits the scope of this refactor.

Judging by its usage in the JavaScript SDK, I expect the following APIs need to be revised:
* purgeInstanceHistory
* * JS example [here](https://github.com/Azure/azure-functions-durable-js/blob/297cc490e70b06eb7ebe418edefb4f5aa2b7fee1/src/durableorchestrationclient.ts#L290)
* raiseEvent
* * JS example [here](https://github.com/Azure/azure-functions-durable-js/blob/297cc490e70b06eb7ebe418edefb4f5aa2b7fee1/src/durableorchestrationclient.ts#L420)
* startNew
* * JS example [here](https://github.com/Azure/azure-functions-durable-js/blob/297cc490e70b06eb7ebe418edefb4f5aa2b7fee1/src/durableorchestrationclient.ts#L674)
* terminate
* * JS example [here](https://github.com/Azure/azure-functions-durable-js/blob/297cc490e70b06eb7ebe418edefb4f5aa2b7fee1/src/durableorchestrationclient.ts#L674)
* getStatusBy
* * JS example [here](https://github.com/Azure/azure-functions-durable-js/blob/297cc490e70b06eb7ebe418edefb4f5aa2b7fee1/src/durableorchestrationclient.ts#L875)

I've omitted the entity APIs that make use of this feature as I believe PowerShell doesn't support those yet. If needed, let me know and I can list those JS examples as well 😄

### LINKED ISSUE #613: Durable functions does not work in combination with App Service AD Authentication / Authorization
Is this a bug or a missing feature?

The observed behavior:
- AD authenticated call returns OrchestrationCheckStatusResponse object
- The provided URLs  in the response object do not work with a bearer token included, or excluded. (404 and 403)
- The Orchestration function does not trigger after Start-NewOrchestration

It works fine locally, and works fine in the cloud when I disable AD authentication/Authorization.

AnatoliB:
Resolves #613
Resolves #597

### DIFF
--- a/src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
+++ b/src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
@@ -51,6 +51,8 @@ function Start-NewOrchestration {
         [object] $DurableClient
     )
 
+    $ErrorActionPreference = 'Stop'
+
     if ($null -eq $DurableClient) {
         $DurableClient = GetDurableClientFromModulePrivateData
     }
--- a/src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
+++ b/src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
@@ -65,7 +65,7 @@ function Start-NewOrchestration {
             "$($DurableClient.rpcBaseUrl)orchestrators/$FunctionName$($InstanceId ? "/$InstanceId" : '')"
         } else {
             # Legacy app frontend path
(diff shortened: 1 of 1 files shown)


Francisco-Gamino (approved):
LGTM

PR #655, opened 2021-07-14
Timestamp snapshot name to use seven digits precision for fraction of a second

### LINKED ISSUE #654: The dependency snapshot name is not unique enough
Currently, a snapshot name contains a timestamp with 3 digits precision for a second. 

Consider increasing the fraction of a second to 7 digits precision.

Francisco-Gamino:
Resolves https://github.com/Azure/azure-functions-powershell-worker/issues/654


AnatoliB (approved):
LGTM

PR #660, opened 2021-08-05
Enable building PowerShell 7.2

Francisco-Gamino:
This PR contains the following:
* Publish a nuget package when merging code to the v4.x/ps7.2 branch
* Update PowerShell worker version to match the Functions Host version
* Update package id to Microsoft.Azure.Functions.PowerShellWorker.PS7.2
* Update worker.config.json to only include PowerShell 7 and 7.2

### DIFF
--- a/PowerShellWorker.Common.props
+++ b/PowerShellWorker.Common.props
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup>
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
-    <Version>3.0.$(BuildNumber)</Version>
+    <Version>4.0.$(BuildNumber)</Version>
     <PackageTags>PowerShell;AzureFunctions;language;Azure;</PackageTags>
-    <PackageLicenseUrl>https://github.com/Azure/azure-functions-powershell-worker/blob/master/LICENSE</PackageLicenseUrl>
+    <PackageLicenseUrl>https://github.com/Azure/azure-functions-powershell-worker/blob/dev/LICENSE</PackageLicenseUrl>
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/Azure/azure-functions-powershell-worker</RepositoryUrl>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


PowerShellWorker.Common.props
```diff
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup>
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
-    <Version>3.0.$(BuildNumber)</Version>
+    <Version>4.0.$(BuildNumber)</Version>
     <PackageTags>PowerShell;AzureFunctions;language;Azure;</PackageTags>
-    <PackageLicenseUrl>https://github.com/Azure/azure-functions-powershell-worker/blob/master/LICENSE</PackageLicenseUrl>
+    <PackageLicenseUrl>https://github.com/Azure/azure-functions-powershell-worker/blob/dev/LICENSE</PackageLicenseUrl>
```
kilasuit, line 6: Not sure this should be or needs to be changed

Francisco-Gamino: @kilasuit -- we no longer have a `master` branch, so when you click at the link above, it redirects you to the default branch, in this case `dev`.

src/worker.config.json
```diff
@@ -4,8 +4,8 @@
         "extensions":[".ps1", ".psm1"],
         "defaultExecutablePath":"dotnet",
         "defaultWorkerPath":"%FUNCTIONS_WORKER_RUNTIME_VERSION%/Microsoft.Azure.Functions.PowerShellWorker.dll",
-        "supportedRuntimeVersions":["6", "7"],
-        "defaultRuntimeVersion":"6",
+        "supportedRuntimeVersions":["7", "7.2"],
+        "defaultRuntimeVersion": null,
         "sanitizeRuntimeVersionRegex":"\\d+"
```
AnatoliB: We need to update this regex, otherwise "7.2" will be interpreted as "7". We need a regex that would map:

- `7` to `7`
- `~7` to `7`
- `7.2` to `7.2`

I think this will work: `"\\d+\\.?\\d*"`

Francisco-Gamino: Good catch. Fixed.

AnatoliB (approved):
LGTM

PR #670, opened 2021-08-20
Preserve requirements.psd1 and use it for managed dependencies snapshot comparison

### LINKED ISSUE #647: When the module version does not *precisely* match the version on the PS Gallery, the worker keeps downloading modules too often
If `requirements.psd1` contains a module version that is similar but is not precisely equal to the version on the PS Gallery, the worker downloads modules on every worker start, consuming more and more storage. For example, this app will run into the issue:

**requirements.psd1**

``` PowerShell
@{ PnP.PowerShell = '1.5' }
```

The problem is that the PS Gallery has version **1.5.0** of this module, but not **1.5**. However, the `Save-Module` cmdlet is "smart" enough to accept that, and it installs version **1.5.0** instead of the specified **1.5**, and this is why the worker does not complain about the `requirements.psd1` content to begin with. However, the next time a worker starts, it checks whether the **1.5** version is present, and it uses a simple string comparison for this, so it does not find it and decides to download dependencies again.

We need to either adjust the module version comparison algorithm to match the `Save-Module` logic and tolerate the trailing ".0", or at least perform this check immediately after the download, so that this module is rejected right away and the user is informed about this.

AnatoliB:
Potentially fixes #647 without the need to introduce a [breaking change](https://github.com/Azure/Azure-Functions/issues/2011).

- When a new snapshot is created, _copy_ the `requirements.psd1` file from the app content to the snapshot directory, so that each snapshot is explicitly associated with the app author intent.
- When the worker evaluates an existing snapshot content to decide whether this snapshot is acceptable or not:
  - Compare the `requirements.psd1` files in the snapshot and in the app content _before_ looking for specific versions.
  - If the `requirements.psd1` content matches, consider this snapshot acceptable (since this snapshot is the result of applying `Save-Module`, regardless of the module version matching rules implemented internally by `Save-Module`).
  - If these files don't match (or the snapshot does not contain `requirements.psd1` because this is an older snapshot, created before rolling out this fix), proceed with looking for specific module versions as usual.

As a result, for example, if `requirements.psd1` contains `'PnP.PowerShell' = '1.5'`, invoking `Save-Module` will still install version **1.5.0**. But, since the original `requirements.psd1` is preserved with the snapshot, the next PowerShell Worker instance will consider this snapshot acceptable and will start using it right away, without creating another one. The problem is solved without introducing a breaking change.

### DIFF
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
             ILogger logger = null)
         {
             _storage = storage ?? new DependencyManagerStorage(GetFunctionAppRootPath(requestMetadataDirectory));
-            _installedDependenciesLocator = installedDependenciesLocator ?? new InstalledDependenciesLocator(_storage);
+            _installedDependenciesLocator = installedDependenciesLocator ?? new InstalledDependenciesLocator(_storage, logger);
             var snapshotContentLogger = new PowerShellModuleSnapshotLogger();
             _installer = installer ?? new DependencySnapshotInstaller(
                                             moduleProvider ?? new PowerShellGalleryModuleProvider(logger),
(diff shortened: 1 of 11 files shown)


Francisco-Gamino (approved):
LGTM. We should add an E2E test case, which can be added in a separate PR.

AnatoliB:
Tested on a private site extension, looks good.

PR #683, opened 2021-09-14
Remove unused extensions.csproj file since E2E tests use extension bundles to register triggers and bindings. 

Francisco-Gamino:


Francisco-Gamino:
Removing installing the extensions via the Core Tools caused the E2E tests to failed in Linux. Closing PR.

PR #720, opened 2021-12-17
Generate SBOM only for release builds

Francisco-Gamino:

Generate SBOM only for release builds.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,12 +23,34 @@ variables:
   buildNumber: $[ counter('build', 400) ] # Start higher than our AppVeyor versions. Every build (pr or branch) will increment.
 
 steps:
+- pwsh: |
+    $isReleaseBuild = $false
+    $releaseBranches = @('v4.x/ps7.2', 'v4.x/ps7.0', 'v3.x/ps7', 'v3.x/ps6', 'v2.x')
+
+    if ($releaseBranches -contains $env:BuildSourceBranchName)
+    {
+        $isReleaseBuild = $true
+    }
+
+    Write-Host "##vso[task.setvariable variable=IsReleaseBuild]$isReleaseBuild"
+    Write-Host "IsReleaseBuild: $isReleaseBuild"
+  displayName: 'Set IsReleaseBuild variable'
+  env:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -23,12 +23,28 @@ variables:
...
       $ErrorActionPreference = "Stop"
-      ./build.ps1 -Clean -Configuration Release -BuildNumber "$(buildNumber)" -AddSBOM -SBOMUtilSASUrl $env:SBOMUtilSASUrl
+      if ($isReleaseBuild)
+      {
+        ./build.ps1 -Clean -Configuration Release -BuildNumber "$(buildNumber)" -AddSBOM -SBOMUtilSASUrl $env:SBOMUtilSASUrl
+      }
+      else
+      {
+        ./build.ps1 -Clean -Configuration Release -BuildNumber "$(buildNumber)"
+      }
```
davidmrdavid, line 47: non-blocking nit: Given the fact that the command we're running here shares the same prefix in both branches, I wonder if you could dynamically construct this command instead.

For example, I'm thinking of something like in the following **pseudocode**.
```
command <- '"./build.ps1 -Clean -Configuration Release -BuildNumber "$(buildNumber)"'
if ($isReleaseBuild)
{
  // add SBOM step
  command <- command  + "-AddSBOM -SBOMUtilSASUrl $env:SBOMUtilSASUrl"
}

Invoke(command);
```

That said, I realize this is just a configuration file so perhaps this is an unnecessary refactoring, so I'm leaving this as a non-blocking suggestion :)

davidmrdavid (approved):
LGTM!

azure-pipelines.yml
Francisco-Gamino: Thanks @davidmrdavid this is a good point. However, I think calling the script directly with the required parameters for each scenario is easier to read and maintain. For example, to debug a scenario where we build a command, we might have to add additional log statements to see what command is ran. If it is ok with you, I will keep it as is.

davidmrdavid: Seems reasonable, merge ahead!

Issue #727, opened 2021-12-24
Context.InstanceId cannot be accessed


anushkakda:
With PowerShell Durable Functions we aren't able to use/pass the instance id in the Orchestrator Function. 
We do see that this is due to the instanceid having an internal access level as defined [here](https://github.com/Azure/azure-functions-powershell-worker/blob/2f47e2e421ba525e30cbdfde734f0a3d94932cc6/src/Durable/OrchestrationContext.cs#L23)

But the instanceId can be accessed with dotnet durable functions. 
![image](https://user-images.githubusercontent.com/40432645/147370549-d90f4f1b-002d-4b22-829f-4e6cb147360d.png)

We would like for the option to be available with PowerShell durable functions as well.

bachuv:
cc @AnatoliB @davidmrdavid

davidmrdavid:
Hi @anushkakda,

Thank you for reaching out. I'll add this to our list of ToDos for triaging in our upcoming meeting, this should easy to patch.

Issue #685, opened 2021-09-17
RaiseEvent-Call doesn't transmit content/body to Orchestrator


amazingidiot:
Hello,
I'm running a durable function with the human interaction pattern like in [the human interaction example](https://github.com/Azure/azure-functions-powershell-worker/blob/dev/examples/durable/DurableApp/HumanInteractionOrchestrator/run.ps1) . I successfully raise the event in a separat azure function via a HTTP call according to the [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-external-events?tabs=powershell). In [line 22](https://github.com/Azure/azure-functions-powershell-worker/blob/df570c205232b9f3865520541ca032e9a5430d1a/examples/durable/DurableApp/HumanInteractionOrchestrator/run.ps1#L22), `$approvalEvent`, the return value of the `Start-DurableExternalEventListener` is used as the input for the activity. I assume that that's the value of the body that's send with the HTTP-call to raise the event, although it is not actually used in the activity itself in the example. I managed to capture the message in the queue, and the body is in there under `TaskMessage.Event.Input`. If I convert the `$approvalEvent` to JSON, it is just `"{}"`.
Is the message body not available in powershell or is there a cmdlet to query the data? The documentation is incomplete on this so far.

Please let me know if I can provide further information! Thank you!

Best regards,
Oliver

amazingidiot:
Hi,
any news on this issue? Do you need more information from me?
Thanks!

davidmrdavid:
Hi @amazingidiot,

No updates yet, sorry to keep you waiting. I've assigned this ticket to me to help track it and I'll start looking into it as soon as possible. It seems to me, from looking at the description, that this is a bug in the SDK implementation, but I'll need a few days of exploration before being able to make that determination. We'll update this thread as soon as possible.

amazingidiot:
Hi @davidmrdavid,
thank you for the reply! Again, if I can provide more information, please tell me.

Best regards!

ghost:
@amazingidiot I was also having the same issue as you so I decided to look in to the issue.

When Wait-DurableTask is used with the -Any parameter a task is returned to PowerShell not the resulting job status, output or event. This is a different behaviour to Wait-DurableTask without the -Any parameter. In this case it returns an array of all the results. This is why you and I were both seeing {} returned.

[...]

PR #746, opened 2022-01-29
Prepare Worker for External DF SDK support (and a few DF bug fixes)

davidmrdavid:
## Problem Statement

The Durable Functions (DF) SDK for PowerShell is currently tightly-coupled with the worker code. This results in two major problems:

1. SDK improvements and patches cannot be released quickly, they're slowed down by the worker-release infrastructure
2. Users cannot choose which version of the SDK they want to use, which makes breaking changes harder to implement.

## Solution

This PR refactors the DF-initialization logic in the worker to allow for users to take a dependency on a DF SDK downloaded from the PowerShell gallery (i.e an _external_ SDK). This sets the foundation to _extract_ the DF SDK from the worker, thereby removing this tight-coupling issue.

In the future, we plan to remove the built-in DF SDK implementation in the worker so that using the external SDK becomes the default option. However, this would be a breaking change and so we have no plans to remove it in the short term.

_Caveat:_ While this PR contains the logic to enable using the external SDK, the specific logic to enable its usage will remain commented out until the external SDK is released on the PowerShell gallery. Still, merging these changes soon will allow us to ensure that the refactoring we've done is safe and doesn't break existing internal-SDK users.

# Other goals

In addition, this PR also fixes several known DF SDK issues. These are listed below:
- https://github.com/Azure/azure-functions-powershell-worker/issues/727
- https://github.com/Azure/azure-functions-powershell-worker/issues/685

These fixes should have probably deserved their own PRs. However, due to time-pressure, they ended up being part of the same branch.

### Changes to enable the use of an external DF SDK

- **The `Durable` directory in the worker has been split into two folders: `DurableSDK` and `DurableWorker`.** The former represents the SDK-specific components that, in the future, could be removed from this repo; this means it also represents the SDK-components that ought to exist in the external SDK. Meanwhile, the `DurableWorker` directory represents DF logic that is intrinsic to the worker, and ought to stay in this repo.

- **The new `DurableSDK` directory in particular is meant to be "drag and dropped" into an external SDK implementation**; it's self-contained. As a result, we have removed any coupling of files within it with utility classes elsewhere in the repo. In practice, this means a few function definitions have been "duplicated" so that they now exist independently inside the "DurableSDK" directory. This is _fine_ since we plan to _eventually_ remove the `DurableSDK` repo from the worker.

- The `OrchestrationInvoker.Invoke` method, where orchestration replay starts, now explicitely checks for whether an external DF SDK is enabled. If that were the case, it defers the _driving_ of orchestration replay to the external SDK. It does this by invoking its "externalInvoker" reference.

- In order to initialize the DF binding uniformly across the external and internal SDKs, we've _delayed_ some of the binding initialization logic. In particular, the operations `pwsh.AddParameter(orchestrationBindingInfo.ParameterName, ...)` and `pwsh.TracePipelineObject` now occur inside the `OrchestrationInvoker` instead of in the Functions-general `Invoke` method. This allows the external SDK's `Invoke` method to control what data it passes to `orchestrationBindingInfo.ParameterName`.

### Other changes

This PR also includes a few urgent bug fixes of the DF SDK. See them below.

- We now expose a `Get-DurableTaskResult` utility which allows users to get the result of already-completed Tasks. This is needed because, if a user didn't immediately bind a task's result to a variable when "awaiting" it, they would previously lose access to it. The code for this was contributed by a community member here: https://github.com/Azure/azure-functions-powershell-worker/pull/753.

- Previously, `WaitForExternalEvent` tasks did not output its results to users. This has been fixed.

- Previously, the fields `InstanceId` and `IsReplaying` where not exposed to users. Now they're available.

## Future plans

After merging and releasing this PR, we'll look to continue the development of the external SDK, add an automated-CI to it, and eventually release it to the PowerShell Gallery.

_co-authored to an equal extent with @michaelpeng36_


src/DurableSDK/Commands/GetDurableTaskResult.cs
```diff
@@ -0,0 +1,36 @@
...
+
+#pragma warning disable 1591 // Missing XML comment for publicly visible type or member 'member'
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable.Commands
+{
+    using System.Collections;
+    using System.Management.Automation;
+    using Microsoft.Azure.Functions.PowerShellWorker.Durable.Tasks;
+
+    [Cmdlet("Get", "DurableTaskResult")]
```
davidmrdavid: This is the community-contributed CmdLet that allows users to get the result of an already-completed Task. In other PLs, users can easily do this by re-await/re-yielding that Task, but the PS programming model doesn't have that. The implementation of this CmdLet isn't as efficient as it could be, but I don't think that's something we can't address after merging it.

src/DurableSDK/Commands/InvokeDurableActivityCommand.cs
```diff
@@ -43,10 +43,8 @@ protected override void EndProcessing()
         {
             var privateData = (Hashtable)MyInvocation.MyCommand.Module.PrivateData;
             var context = (OrchestrationContext)privateData[SetFunctionInvocationContextCommand.ContextKey];
-            var loadedFunctions = FunctionLoader.GetLoadedFunctions();
 
             var task = new ActivityInvocationTask(FunctionName, Input, RetryOptions);
-            ActivityInvocationTask.ValidateTask(task, loadedFunctions);
```
davidmrdavid, line 49: These two were removed since they're incompatible with the external SDK. In addition, other DF SDKs don't perform this validation steps. If we want them, they ought to exist outside the DF SDKs.

I do agree that these are useful validations to have though, but if we perform these checks in the DF SDK components, then we're creating a strong dependency with worker-specific utilities.

src/DurableSDK/DurableTaskHandler.cs
```diff
@@ -57,6 +60,13 @@ internal class DurableTaskHandler
                                 output(eventResult);
                             }
                             break;
+                        case HistoryEventType.EventRaised:
+                            var eventRaisedResult = GetEventResult(completedHistoryEvent);
+                            if (eventRaisedResult != null)
+                            {
+                                output(eventRaisedResult);
+                            }
+                            break;
```
davidmrdavid: This allows `WaitForExternalEvent` tasks to output a result

src/DurableSDK/DurableTaskHandler.cs
```diff
@@ -195,6 +209,21 @@ internal class DurableTaskHandler
...
+            foreach (var task in tasksToQueryResultFor) {
+                var scheduledHistoryEvent = task.GetScheduledHistoryEvent(context, true);
+                var processedHistoryEvent = task.GetCompletedHistoryEvent(context, scheduledHistoryEvent, true);
+                if (processedHistoryEvent != null)
+                {
+                    output(GetEventResult(processedHistoryEvent));
+                }
+            }
+        }
+
```
davidmrdavid: Part of the community-contributed logic for "Get-TaskResult". Again, this could use some caching, but I'd rather optimize that later since this is feature is **urgently** missing.

src/DurableSDK/DurableTaskHandler.cs
```diff
@@ -206,15 +235,41 @@ private static object GetEventResult(HistoryEvent historyEvent)
...
             }
             else if (historyEvent.EventType == HistoryEventType.EventRaised)
             {
-                return TypeExtensions.ConvertFromJson(historyEvent.Input);
+                return ConvertFromJson(historyEvent.Input);
             }
             return null;
         }
 
+        public static object ConvertFromJson(string json)
```
davidmrdavid, line 246: This used to exist in `TypeExtensions`, creating a tight-dependency with worker utilities. It's been copied here to allow this folder to be readily copied out into an external SDK

src/DurableSDK/Commands/GetDurableTaskResult.cs
AnatoliB: This addition seems to be independent from the main purpose of the PR. If this is not very difficult, I recommend separating this into its own PR. Unless this is indeed too difficult.

src/DurableSDK/Commands/InvokeDurableActivityCommand.cs
AnatoliB: The intent of `ActivityInvocationTask.ValidateTask` is to provide user-friendly error messages when the specified activity function does not exits (for example, because of a typo in the name). Without this code, the error messages are really confusing and unhelpful. If you remove it, we need an adequate replacement.

davidmrdavid:
Closing in favor of https://github.com/Azure/azure-functions-powershell-worker/pull/839

PR #753, opened 2022-02-04
Allow completed Tasks to expose their results

### LINKED ISSUE #685: RaiseEvent-Call doesn't transmit content/body to Orchestrator

davidmrdavid:
Contribution by @oobegreg, thank you!

resolves #685

Currently, users do not have a way of obtaining the result of a DF Task that has already completed. This means that if a user did not assign a Task onto a variable when first invoking it, then they have no way of accessing its result. 

This is an urgently missing feature and @oobegreg is proposing a CmdLet named `Get-DurableTaskResult` that allows us to do just that.

In general, this serves the purpose of performing an `await` (or `yield` in Python and JS)  on a completed Durable Task, because in those languages re-awaiting a completed Task is a supported means of extracting that Task's result. Another way to do this is to expose a `.result` property on the `DurableTask` class and have users access that. However, I think this CmdLet-based approach is also sound, and I don't see why we couldn't have both given that users in other PLs can both access their classes' `.result` property *and* call `yield/await` for the same effect.

My recommendation is to merge this contribution, but I would like to get everyone's feedback first. Thanks!

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/Durable/Commands/GetDurableTaskResult.cs
+++ b/src/Durable/Commands/GetDurableTaskResult.cs
@@ -1,0 +1,36 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#pragma warning disable 1591 // Missing XML comment for publicly visible type or member 'member'
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable.Commands
+{
+    using System.Collections;
+    using System.Management.Automation;
+    using Microsoft.Azure.Functions.PowerShellWorker.Durable.Tasks;
+
+    [Cmdlet("Get", "DurableTaskResult")]
+    public class GetDurableTaskResultCommand : PSCmdlet
+    {
+        [Parameter(Mandatory = true)]
+        [ValidateNotNull]
+        public DurableTask[] Task { get; set; }
+
+        private readonly DurableTaskHandler _durableTaskHandler = new DurableTaskHandler();
+
+        protected override void EndProcessing()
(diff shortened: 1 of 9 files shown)


AnatoliB (commented):
Looks good, thank you! I would also love to see some E2E test coverage for the new command.

davidmrdavid:
I'll be merging this into a separate branch that I own, so that I can add the E2E tests

PR #758, opened 2022-02-09
Switch from Grpc.Core to Grpc.Net.Client

### LINKED ISSUE #751: Switch to new gRPC package
Per [this readme](https://github.com/grpc/grpc/tree/master/src/csharp#grpc-c) and [this blogpost](https://grpc.io/blog/grpc-csharp-future/), the `Grpc.Net.Client` nuget package is now the recommended gRPC implementation for .NET/C# as opposed to the `Grpc.Core` package. The old package is currently in maintenance mode and will be deprecated in May 2022

michaelpeng36:

resolves #751 to facilitate M1 compatibility for the PowerShell worker V4 using PowerShell 7.2.

### Pull request checklist

### DIFF
--- a/src/Messaging/MessagingStream.cs
+++ b/src/Messaging/MessagingStream.cs
@@ -3,9 +3,11 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
+using Grpc.Core;
 using Grpc.Net.Client;
 using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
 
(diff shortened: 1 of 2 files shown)


michaelpeng36:
I currently plan to backport this change to `v4.x/ps7.0` since `Grpc.Core` is to be deprecated in May 2022, and I anticipate we will still be doing releases of the PowerShell Worker at that point.

michaelpeng36:
The changes in this PR are based on the changes implemented in the .NET worker (see [this](https://github.com/Azure/azure-functions-dotnet-worker/blob/3e65d91314fc0920349ecd5eaadc36162be868f3/src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs#L58) for the specific file). One difference here, however, is that `grpcUri` is a nullable `Uri` in the .NET worker. Do you think that is necessary here?

ejizba (approved):

Francisco-Gamino (approved):
LGTM

Francisco-Gamino:
> The changes in this PR are based on the changes implemented in the .NET worker (see [this](https://github.com/Azure/azure-functions-dotnet-worker/blob/3e65d91314fc0920349ecd5eaadc36162be868f3/src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs#L58) for the specific file). One difference here, however, is that `grpcUri` is a nullable `Uri` in the .NET worker. Do you think that is necessary here?

I do not think so. You're current implementation works.

PR #786, opened 2022-04-05
Add Get-TaskResult CmdLet for DF

davidmrdavid:

Related to https://github.com/Azure/azure-functions-powershell-worker/issues/685 and https://github.com/Azure/azure-functions-powershell-worker/pull/753

Currently, users do not have a way of obtaining the result of a DF Task that has already completed. This means that if a user did not assign a Task onto a variable when first invoking it, then they have no way of accessing its result.

This is an urgently missing feature and user oobegreg is proposing a CmdLet named Get-DurableTaskResult that allows us to do just that.

In general, this serves the purpose of performing an await (or yield in Python and JS) on a completed Durable Task, because in those languages re-awaiting a completed Task is a supported means of extracting that Task's result. Another way to do this is to expose a .result property on the DurableTask class and have users access that. However, I think this CmdLet-based approach is also sound, and I don't see why we couldn't have both given that users in other PLs can both access their classes' .result property and call yield/await for the same effect.

### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR: https://github.com/Azure/azure-functions-powershell-worker/issues/797
  * [x] Otherwise: I've added my notes to `release_notes.md`
  * [ ] Otherwise: Backport tracked by issue/PR https://github.com/Azure/azure-functions-powershell-worker/pull/799, https://github.com/Azure/azure-functions-powershell-worker/pull/798

### Additional information

This needs a documentation update (to announce the new CmdLet) and also needs to be backported to all DF-supported branches.

Where should this documentation issue requested by the PR checklist be created? In the azure docs repo? I suppose that we can create it in this repo as well, just checking for this group goes about it :-)

### DIFF
--- a/src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
+++ b/src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
@@ -235,6 +235,8 @@ function New-DurableOrchestrationCheckStatusResponse {
     The TaskHubName of the orchestration instance that will handle the external event.
 .PARAMETER ConnectionName
     The name of the connection string associated with TaskHubName
+.PARAMETER AppCode
+    The Azure Functions system key
 #>
 function Send-DurableExternalEvent {
     [CmdletBinding()]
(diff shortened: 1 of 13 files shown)


AnatoliB:
@davidmrdavid Please add a work item to document the new cmdlet.

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
```diff
@@ -235,6 +235,8 @@ function New-DurableOrchestrationCheckStatusResponse {
     The TaskHubName of the orchestration instance that will handle the external event.
 .PARAMETER ConnectionName
     The name of the connection string associated with TaskHubName
+.PARAMETER AppCode
```
AnatoliB, line 246: These changes do not seem to be related to the declared PR goal. I will not be insisting on necessarily separating them, but I want us to be mindful of this.

AnatoliB (approved):

davidmrdavid:
Added documentation issue here: https://github.com/Azure/azure-functions-powershell-worker/issues/797

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
davidmrdavid: Good call,  I forgot about this addition. At the very least, I should have called this out. I'll keep an eye out in the future

PR #793, opened 2022-04-07
Rename DF components

davidmrdavid:
This PR contains some of the changes originally proposed in https://github.com/Azure/azure-functions-powershell-worker/pull/746

This PR introduces two **non-behavioral** changes (i.e the end-user should not be affected in any way):

- The `Durable` directory in the worker has been split into two folders: `DurableSDK` and `DurableWorker`. The former represents the SDK-specific components that, in the future, could be removed from this repo; this means it also represents the SDK-components that ought to exist in the external SDK. Meanwhile, the `DurableWorker` directory represents DF logic that is intrinsic to the worker, and ought to stay in this repo.
- The new `DurableSDK` directory in particular is meant to be "drag and dropped" into an external SDK implementation; it's self-contained. As a result, we have removed any coupling of files within it with utility classes elsewhere in the repo. In practice, this means a few function definitions have been "duplicated" so that they now exist independently inside the "DurableSDK" directory. This is fine since we plan to eventually remove the `DurableSDK` repo from the worker.

In other words, I'm just slightly re-organizing the codebase to set the stage for the external SDK work. I decided to make this its own separate PR because a lot of files are affected and yet no functionality is changed, so I wanted to make sure this did not **distract** from the future PR where we actually change some behavior to enable the external DF SDK.

### Pull request checklist

  * [x] Otherwise: Backport tracked by issue/PR #858, #859

### Additional information

This needs to be backported to all DF-supported branches.

co-authored to an equal extent with @michaelpeng36


release_notes.md
```diff
@@ -1 +1 @@
-* Bug fix: Activity Functions can now use output bindings (https://github.com/Azure/azure-functions-powershell-worker/issues/646)
\ No newline at end of file
+* New feature: The `Get-TaskResult` CmdLet can now be used to get the result of already completed DF Tasks! Thanks @oobegreg for your contribution!
```
AnatoliB: This belongs to a different PR, right?

test/E2E/TestFunctionApp/profile.ps1
```diff
@@ -0,0 +1,22 @@
+# Azure Functions profile.ps1
```
AnatoliB: Is this auto-generated? Unless you really need it, I suggest you exclude it from the PR

AnatoliB (approved):
LGTM (just exclude profile.ps1 if not needed)

test/E2E/TestFunctionApp/profile.ps1
davidmrdavid: I think it was. Will remove

release_notes.md
davidmrdavid: Yes, I believe so. This diff should disappear once the other PR is merged.

davidmrdavid: Yes, believe so. I think this diff will disappear once the PR for that feature is merged. I'll wait for that.

PR #839, opened 2022-07-12
Enable external DF SDK

davidmrdavid:
This PR is the culmination of several smaller PRs that originated here: https://github.com/Azure/azure-functions-powershell-worker/pull/746

This PR changes the Durable Functions logic in the worker to account for the possibility of an external Durable Functions SDK implementation. That SDK would be installed as a managed dependency, from the PowerShell Gallery.

If this external SDK were detected, we defer the invocation of **orchestrator** Function types to that external SDK. All other invocation types are still handled by the worker. In this refactoring, I've also simplified some of the orchestration-detection logic as shown in this PR's `InvokeFunction` method.

This PR is in collaboration with @michaelpeng36

### Pull request checklist

### Additional information

Before merging this PR, we will need to decide on the namespace of the new external Durable Functions SDK. In fact, it may be best to fully reserve it prior to merging.


src/DurableSDK/Commands/InvokeDurableActivityCommand.cs
```diff
@@ -43,10 +43,8 @@ protected override void EndProcessing()
         {
             var privateData = (Hashtable)MyInvocation.MyCommand.Module.PrivateData;
             var context = (OrchestrationContext)privateData[SetFunctionInvocationContextCommand.ContextKey];
-            var loadedFunctions = FunctionLoader.GetLoadedFunctions();
 
             var task = new ActivityInvocationTask(FunctionName, Input, RetryOptions);
-            ActivityInvocationTask.ValidateTask(task, loadedFunctions);
```
davidmrdavid: Note to self: need to re-enable this

src/DurableSDK/DurableTaskHandler.cs
```diff
@@ -83,7 +85,7 @@ internal class DurableTaskHandler
                                         retryOptions.MaxNumberOfAttempts,
                                         onSuccess:
                                             result => {
-                                                output(TypeExtensions.ConvertFromJson(result));
```
davidmrdavid: Throughout this file, we remove the dependency from `TypeExtensions`. This allows us to remove the tight-dependency between the `DurableSDK` component, and the worker.

In general, the previous state of the Durable logic in the worker was tightly coupled with worker-specific functionality. By removing this tight-coupling, it becomes easier to what pieces of DF logic need to exist in the worker, and which pieces can exist in the separate SDK.  Looking towards the future, it also makes it easier to identify what code would be safe to remove from the worker once the DF SDK exists in an external package: the code in the `DurableSDK` namespace should be safe to remove after some time.

src/DurableSDK/DurableTaskHandler.cs
```diff
@@ -232,15 +234,41 @@ private static object GetEventResult(HistoryEvent historyEvent)
...
             }
             else if (historyEvent.EventType == HistoryEventType.EventRaised)
             {
-                return TypeExtensions.ConvertFromJson(historyEvent.Input);
+                return ConvertFromJson(historyEvent.Input);
             }
             return null;
         }
 
+        public static object ConvertFromJson(string json)
```
davidmrdavid: This used to exist in TypeExtensions, creating a tight-dependency with worker utilities. It's been copied here to allow this folder to be readily copied out into an external SDK

src/DurableSDK/ExternalInvoker.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System;
+    using System.Collections;
+    using System.Management.Automation;
+
+    internal class ExternalInvoker : IExternalInvoker
```
davidmrdavid: This class just defines the interface we expect in the external SDK's orchestration invoker. It's here mostly to help document the contract between the worker and the external SDK, and to provide some amount of type-checking.

src/DurableSDK/IExternalInvoker.cs
```diff
@@ -0,0 +1,16 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
+{
+    using System.Collections;
+
+    // Represents a contract for the
```
davidmrdavid: note to self: this comment is incomplete

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -11,58 +11,98 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
     using System.Linq;
     using System.Management.Automation;
 
-    using PowerShellWorker.Utility;
     using Microsoft.Azure.Functions.PowerShellWorker.Durable.Actions;
 
     internal class OrchestrationInvoker : IOrchestrationInvoker
     {
-        public Hashtable Invoke(OrchestrationBindingInfo orchestrationBindingInfo, IPowerShellServices pwsh)
+        private IExternalInvoker _externalInvoker;
```
davidmrdavid: note to self: do we really want this underscore?

src/DurableSDK/IPowerShellServices.cs
```diff
@@ -5,17 +5,26 @@
...
 {
+    using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
     using System;
     using System.Management.Automation;
 
     internal interface IPowerShellServices
     {
+        PowerShell GetPowerShell();
+
+        bool UseExternalDurableSDK();
```
davidmrdavid: note to self: I think this method can be renamed to something like "HasExternalDurableSDK()"

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -11,58 +11,98 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
...
             }
-            finally
+        }
+
+        public static object CreateReturnValueFromFunctionOutput(IList<object> pipelineItems)
+        {
+            if (pipelineItems == null || pipelineItems.Count <= 0)
             {
-                pwsh.ClearStreamsAndCommands();
+                return null;
```
davidmrdavid, line 101: note to self: this change doesn't seem right

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -72,7 +112,12 @@ public Hashtable Invoke(OrchestrationBindingInfo orchestrationBindingInfo, IPowe
             object customStatus)
         {
             var orchestrationMessage = new OrchestrationMessage(isDone, actions, output, customStatus);
-            return new Hashtable { { AzFunctionInfo.DollarReturn, orchestrationMessage } };
+            return new Hashtable { { "$return", orchestrationMessage } };
```
davidmrdavid: note to self: looks like we should convert `"$return"` back to `AzFunctionInfo.DollarReturn`

src/DurableSDK/PowerShellExtensions.cs
```diff
@@ -0,0 +1,69 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System.Collections;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
```
davidmrdavid: Same as we did for `TypeExtensions`, I want to remove the dependencies of the `DurableSDK` to the worker. So here I copy over the PS-Extensions logic.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,144 @@
...
+    using Microsoft.Azure.Functions.PowerShellWorker.Utility;
+    using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
+    using Newtonsoft.Json;
 
     internal class PowerShellServices : IPowerShellServices
     {
-        private const string SetFunctionInvocationContextCommand =
-            "Microsoft.Azure.Functions.PowerShellWorker\\Set-FunctionInvocationContext";
+        private readonly string SetFunctionInvocationContextCommand;
+        private const string ExternalDurableSDKName = "DurableSDK";
```
davidmrdavid: heads up: before merging, we need a proper namespace reserved for the external SDK.

src/DurableSDK/DurableTaskHandler.cs
```diff
@@ -232,15 +234,41 @@ private static object GetEventResult(HistoryEvent historyEvent)
...
+        {
+            object retObj = JsonObject.ConvertFromJson(json, returnHashtable: true, error: out _);
```
Francisco-Gamino: Do we care about depth? If we do, we should set the depth.

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -11,58 +11,98 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
...
+            OrchestrationBindingInfo orchestrationBindingInfo,
+            IPowerShellServices powerShellServices)
         {
             try
             {
-                var outputBuffer = new PSDataCollection<object>();
-                var context = orchestrationBindingInfo.Context;
+                if (powerShellServices.UseExternalDurableSDK())
+                {
+                    return InvokeExternalDurableSDK(powerShellServices);
```
Francisco-Gamino, line 29: Could you please share the link to the `externalInvoker.Invoke` code?

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -11,58 +11,98 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
...
 
-                if (shouldStop)
+            var (shouldStop, actions) =
+                orchestrationBindingInfo.Context.OrchestrationActionCollector.WaitForActions(asyncResult.AsyncWaitHandle);
+
+            if (shouldStop)
+            {
+                // The orchestration function should be stopped and restarted
+                powerShellServices.StopInvoke();
+                // return (Hashtable)orchestrationBindingInfo.Context.OrchestrationActionCollector.output;
```
Francisco-Gamino: If this is not needed, please remove it.

src/DurableSDK/OrchestrationInvoker.cs
Francisco-Gamino: Do you have unit test that covers this code path?

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,144 @@
...
+                // the Import-Module statement and we should throw an Exception.
+                // Otherwise, we use the InternalDurableSDK
+                var availableModules = pwsh.AddCommand(Utils.GetModuleCmdletInfo)
+                    .AddParameter("Name", ExternalDurableSDKName)
+                    .InvokeAndClearCommands<PSModuleInfo>();
+                if (availableModules.Count() > 0)
+                {
+                    // TODO: evaluate if there is a better error message or exception type to be throwing.
+                    // Ideally, this should never happen.
+                    throw new InvalidOperationException("The external Durable SDK was detected, but unable to be imported.", e);
```
Francisco-Gamino: What about this error message? `Failed to import the <ModuleName> module. Please make sure this module is available in the PowerShell worker path. For instructions on how to install this module, please see <link to fwlink>`.  

Also, this should be a string in a resource file.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,144 @@
...
+                }
+                _useExternalDurableSDK = false;
+            }
+            //_useExternalDurableSDK = false;
```
Francisco-Gamino: Please remove this if it is not needed.

src/Utility/TypeExtensions.cs
```diff
@@ -142,7 +142,7 @@ public static object ConvertFromJson(string json)
         private static string ConvertToJson(object fromObj)
         {
             var context = new JsonObject.ConvertToJsonContext(
-                maxDepth: 4,
+                maxDepth: 10, //TODO: fix
```
Francisco-Gamino: Could you please add a comment on why were are using `maxDepth = 10`?

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,144 @@
...
+        private const string InternalDurableSDKName = "Microsoft.Azure.Functions.PowerShellWorker";
 
         private readonly PowerShell _pwsh;
-        private bool _hasSetOrchestrationContext = false;
+        private bool _hasInitializedDurableFunction = false;
+        private readonly bool _useExternalDurableSDK = false;
 
         public PowerShellServices(PowerShell pwsh)
         {
+            //This logic will be commented out until the external SDK is published on the PS Gallery
```
davidmrdavid: note to self: remove this comment

src/DurableSDK/PowerShellServices.cs
davidmrdavid: note to self: remove

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,144 @@
...
+                JsonConvert.DeserializeObject<OrchestrationContext>(context.Data.String));
+
+            if (_useExternalDurableSDK)
+            {
+                Collection<Func<PowerShell, object>> output = _pwsh.AddCommand(SetFunctionInvocationContextCommand)
+                    // The external SetFunctionInvocationContextCommand expects a .json string to deserialize
+                    // and writes an invoker function to the output pipeline.
+                    .AddParameter("OrchestrationContext", context.Data.String)
+                    .InvokeAndClearCommands<Func<PowerShell, object>>();
+                if (output.Count() == 1)
```
davidmrdavid: note to self: rename this operation to make it more readable

src/Modules/Microsoft.Azure.Functions.PowerShellWorker/Microsoft.Azure.Functions.PowerShellWorker.psm1
```diff
@@ -11,7 +11,7 @@ Set-Alias -Name Start-NewOrchestration -Value Start-DurableOrchestration
 
 function GetDurableClientFromModulePrivateData {
     $PrivateData = $PSCmdlet.MyInvocation.MyCommand.Module.PrivateData
-    if ($PrivateData -eq $null -or $PrivateData['DurableClient'] -eq $null) {
+    if ($null -eq $PrivateData -or $null -eq $PrivateData['DurableClient']) {
```
davidmrdavid: note to self: reorder this check

src/PowerShell/PowerShellManager.cs
```diff
@@ -257,7 +263,7 @@ internal void InvokeProfile(string profilePath)
         private void SetInputBindingParameterValues(
             AzFunctionInfo functionInfo,
             IEnumerable<ParameterBinding> inputData,
-            DurableController durableController,
+            string orchParamName,
```
davidmrdavid: note to self: consider fully-expanding this variable name

src/Utility/TypeExtensions.cs
davidmrdavid: note to self: add this in another PR?

test/E2E/TestFunctionApp/profile.ps1
```diff
@@ -0,0 +1,22 @@
+# Azure Functions profile.ps1
```
davidmrdavid: I think this file can be removed

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -11,58 +11,98 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
...
+
+        public Hashtable Invoke(
+            OrchestrationBindingInfo orchestrationBindingInfo,
+            IPowerShellServices powerShellServices)
         {
             try
             {
-                var outputBuffer = new PSDataCollection<object>();
-                var context = orchestrationBindingInfo.Context;
+                if (powerShellServices.HasExternalDurableSDK())
```
davidmrdavid: Unfortunately, the diff in this section of the code makes the edit seem more complicated than it is.
In short, the **previous** logic of the `Invoke` method in here is now contained within the `InvokeInternalDurableSDK` . However, for the rest of this file, the diff will show parts of the old method interleaved in weird ways with new methods.

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -11,58 +11,98 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
...
             }
-            finally
+        }
+
+        public static object CreateReturnValueFromFunctionOutput(IList<object> pipelineItems)
+        {
+            if (pipelineItems == null || pipelineItems.Count <= 0)
             {
-                pwsh.ClearStreamsAndCommands();
+                return null;
```
davidmrdavid, line 101: Just for clarity: the `pwsh.ClearStreamsAndCommand()` call here didn't use to be  in the `CreateReturnValueFromFunctionOutput` method. The implementation of this method has not changed at all, it's just that the diff is acting up and still showing code from the previous `Invoke` implementation.

src/DurableWorker/DurableController.cs
```diff
@@ -88,46 +94,22 @@ public bool TryGetInputBindingParameterValue(string bindingName, out object valu
...
-            // "Orchestrator functions should never use any input or output bindings other than the orchestration trigger binding.
-            //  Doing so has the potential to cause problems with the Durable Task extension because those bindings may not obey the single-threading and I/O rules."
-            //
-            // Therefore, it's by design that input data contains only one item, which is the metadata of the orchestration context.
-            var context = inputData[0];
-
-            return new OrchestrationBindingInfo(
-                context.Name,
-                JsonConvert.DeserializeObject<OrchestrationContext>(context.Data.String));
-        }
```
davidmrdavid, line 131: this method is removed because it's logic has been spread out to account for the external SDK

src/Utility/TypeExtensions.cs
```diff
@@ -142,7 +142,7 @@ public static object ConvertFromJson(string json)
         private static string ConvertToJson(object fromObj)
         {
             var context = new JsonObject.ConvertToJsonContext(
-                maxDepth: 4,
+                maxDepth: 50,
```
davidmrdavid: We need this change because the "fast" Durable Functions OOProc model utilizes deeply nested JSON objects to represented WhenAll and WhenAny actions. I'm happy to discuss just how deep we really need this to be, but 4 is certainly too low for even the simplest WhenAny/WhenAll JSON.

test/Unit/Durable/DurableControllerTests.cs
```diff
@@ -133,70 +136,52 @@ public void TryGetInputBindingParameterValue_RetrievesOrchestrationContextParame
...
-            var durableController = CreateDurableController(durableFunctionType);
-
-            _mockPowerShellServices.Setup(_ => _.SetOrchestrationContext(It.IsAny<OrchestrationContext>()));
-            durableController.BeforeFunctionInvocation(inputData);
-
-            var invoked = durableController.TryInvokeOrchestrationFunction(out var actualResult);
-            Assert.False(invoked);
-            Assert.Null(actualResult);
-        }
-
```
davidmrdavid: We don't have this method anymore because of how the logic has been refactored. We now guarantee that the orchestration invoker is only called for orchestrator types, instead of always calling it and requiring `TryInvokeOrchestratorFunction` to do the check for us.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,146 @@
...
+                if (availableModules.Count() > 0)
+                {
+                    // TODO: evaluate if there is a better error message or exception type to be throwing.
+                    // Ideally, this should never happen.
+                    throw new InvalidOperationException("The external Durable SDK was detected, but unable to be imported.", e);
+                }
+                hasExternalDurableSDK = false;
+            }
+
+            if (hasExternalDurableSDK)
```
Francisco-Gamino: Do we log anywhere what Durable SDK (internal or external) version the PowerShell worker is using?

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,146 @@
...
+                // If more than 1 element is present in the output pipeline, we cannot trust that we have
+                // obtained the external orchestrator invoker; i.e the output contract is not met.
+                var outputContractIsMet = output.Count() == 1;
+                if (outputContractIsMet)
+                {
+                    externalInvoker = new ExternalInvoker(output[0]);
+                }
+                else
+                {
+                    throw new InvalidOperationException($"Only a single output was expected for an invocation of {SetFunctionInvocationContextCommand}");
```
Francisco-Gamino: Please add string resources for all messages that are displayed to the user.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,146 @@
...
             {
                 _pwsh.AddCommand(SetFunctionInvocationContextCommand)
                     .AddParameter("Clear", true)
                     .InvokeAndClearCommands();
             }
         }
 
+        public void TracePipelineObject()
+        {
+            _pwsh.AddCommand("Microsoft.Azure.Functions.PowerShellWorker\\Trace-PipelineObject");
```
Francisco-Gamino: Should we create a file with constants (for the cmdlet names) which can be used across the whole project?

src/Utility/TypeExtensions.cs
Francisco-Gamino: Could you please add a comment on what 50 means in this context?

Francisco-Gamino (commented):
Thank you @davidmrdavid for sending this out! I left a few comments. Thank you.

src/DurableSDK/OrchestrationInvoker.cs
davidmrdavid: Removed, thanks for catching this

src/DurableSDK/PowerShellServices.cs
davidmrdavid: I changed this to import from the string-resource file. However, I wasn't sure what link to give them for suggestions. Where you thinking of something specific or just a standard PS debugging document? I figured we could link them this: https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-powershell?tabs=portal#dependency-management

src/DurableSDK/PowerShellServices.cs
davidmrdavid: It's been removed

src/Utility/TypeExtensions.cs
davidmrdavid: I added a comment. The gist of it is that Durable Functions, irrespective of the new external package we're developing, may produce deeply nested JSON objects. Therefore, a depth of 4 was fairly limited and error prone for durable users.

src/DurableSDK/PowerShellServices.cs
davidmrdavid: Good idea! Done

src/DurableSDK/PowerShellServices.cs
davidmrdavid: Added!

src/Utility/TypeExtensions.cs
davidmrdavid: Copy-pasting from a related comment: 

I added a comment. The gist of it is that Durable Functions, irrespective of the new external package we're developing, may produce deeply nested JSON objects. Therefore, a depth of 4 was fairly limited and error prone for durable users.

davidmrdavid: I chose 50 arbitrarily. I just wanted to make it high enough to cover most durable cases. In theory, I'd be ok with setting it even higher.

src/DurableSDK/PowerShellServices.cs
davidmrdavid: I added a log for this in the PowerShellManager.InvokeFunction method

src/DurableSDK/OrchestrationInvoker.cs
davidmrdavid: Yes, although that code is not ready for review. Still, if you're interested, it's here: https://github.com/Azure/azure-functions-durable-powershell-private/blob/cd02ada03f353ccc1ba0cb4f87d6f54243805129/samples/durableApp/Modules/DurableSDK/DurableEngine/OrchestrationInvoker.cs#L61

src/Utility/TypeExtensions.cs
Francisco-Gamino: Make sense, thank you for the detailed explanation.

src/Utility/TypeExtensions.cs
```diff
@@ -141,8 +141,13 @@ public static object ConvertFromJson(string json)
...
         private static string ConvertToJson(object fromObj)
         {
+            /* we set the max-depth to 50 because the Durable Functions Extension
+             * may produce deeply nested JSON-Objects when callig its
+             * WhenAll/WhenAny APIs. The value 50 is arbitrarily chosen to be
+             * "deep enough" for the vast majority of cases.
+             */
             var context = new JsonObject.ConvertToJsonContext(
-                maxDepth: 4,
+                maxDepth: 50,
```
AnatoliB: This looks like a _potential_ breaking change. Please make sure you check all the ConvertToJson use cases.

src/DurableSDK/OrchestrationInvoker.cs
```diff
@@ -11,58 +11,97 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
...
     using System.Management.Automation;
 
-    using PowerShellWorker.Utility;
     using Microsoft.Azure.Functions.PowerShellWorker.Durable.Actions;
 
     internal class OrchestrationInvoker : IOrchestrationInvoker
     {
-        public Hashtable Invoke(OrchestrationBindingInfo orchestrationBindingInfo, IPowerShellServices pwsh)
+        private IExternalOrchestrationInvoker externalInvoker;
+        internal static string isOrchestrationFailureKey = "IsOrchestrationFailure";
```
AnatoliB: Wondering why not `const` (which would express the idea better, and would probably be slightly efficient than `static`)

davidmrdavid: Yeah this should be `const`, I missed that during development.
Actually, I've moved this to our constant strings resource file, and changed all references to just read this value from there. Let me know if that seems reasonable to you!

src/Utility/TypeExtensions.cs
davidmrdavid: Yes, I was thinking about that. I could look to make this change _only_ for the Durable Functions bindings. Let me consider this a little longer.

davidmrdavid:
I believe I have addressed most (if not all) the feedback, so I'll be requesting another round of reviews :) . Thanks in advance

src/DurableSDK/OrchestrationInvoker.cs
AnatoliB: No big deal either way, but I would generally use string resources mostly for messages that may need to be localized, rephrased, etc. without changing functionality. String identifiers required for your code _correctness_ (is this the case here?) should stay in the code. This is arguable, so I will not insist.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,140 @@
...
         {
-            _pwsh = pwsh;
+            // We attempt to import the external SDK upon construction of the PowerShellServices object.
+            // We maintain the boolean member hasExternalDurableSDK in this object rather than
+            // DurableController because the expected input and functionality of SetFunctionInvocationContextCommand
+            // may differ between the internal and external implementations.
+
+            try
+            {
+                pwsh.AddCommand(Utils.ImportModuleCmdletInfo)
```
Francisco-Gamino: For this scenario, what if we did the following?

1) Do a `Get-Module -list PowerShellWorkerStrings.ExternalDurableSDKName` and log the module name and version. This will tell us if the module is the worker path. Additionally, you could track or build a detector which will tell you what `PowerShellWorkerStrings.ExternalDurableSDKName` version is currently being used by our customers. 
2) If the output of (1) is not null, then import the module. Otherwise, write the log with guidance to the user on how to install the module, e.g., `Please make sure this module is available in the PowerShell worker path. For instructions on how to install this module, please see '{1}'.`
3) If the module could not be imported, get the error details using the `Get-Error` cmdlet and log it (to Kusto) and display the error `exceptionMessage` to the user. Follow up question: Are these user logs, Kusto logs or both?

davidmrdavid:
Thanks @Francisco-Gamino, for the suggestions. I've incorporated them all.

In particular, here's the latest changes.
1) The enablement of the external DF SDK is now guarded by a feature flag. This feature flag is the environment variable: `ExternalDurablePowerShellSDK`.
2) If the feature flag is enabled, we call `tryEnablingExternalDurableSDK()` in the `DurableController`. Otherwise, we do not and therefore default to the built-in logic that we use today. You can see this here:

https://github.com/Azure/azure-functions-powershell-worker/blob/9ab5b097e45f4f55f10d0339f809092e54160beb/src/DurableWorker/DurableController.cs#L62-L70

3) If `tryEnablingExternalDurableSDK()` is called, we begin the process by using `Get-Module` to see if the fully qualified external DF SDK can be found in the worker path. If we cannot find it, we log to Kusto (and the user) that we did not find the module. You can see this here:

https://github.com/Azure/azure-functions-powershell-worker/blob/9ab5b097e45f4f55f10d0339f809092e54160beb/src/DurableSDK/PowerShellServices.cs#L79-L101

3.1) If the module was found, we then try to import it in the function `tryImportingDurableSDK()`. In it, we use `Import-Module` to try and load the DF SDK, and report the module version that was loaded as a log to the user and kusto. If it failed, we try to use `Get-Error` to log the error. Otherwise, we log the C#-level exception we caught. All this logic is here: https://github.com/Azure/azure-functions-powershell-worker/blob/9ab5b097e45f4f55f10d0339f809092e54160beb/src/DurableSDK/PowerShellServices.cs#L28-L75

Finally, I added some minimal unit tests to check that the new code paths are only accessible through the feature flag. You can see these here:

https://github.com/Azure/azure-functions-powershell-worker/blob/9ab5b097e45f4f55f10d0339f809092e54160beb/test/Unit/Durable/DurableControllerTests.cs#L253-L319

I'll be further validating this manually in the next couple of hours, but I would still appreciate to get your thoughts on these changes assuming my manual validation were to succeed. Thanks!

FYI @michaelpeng36

src/DurableSDK/PowerShellServices.cs
davidmrdavid: Thanks @Francisco-Gamino. I've incorporated most of this feedback, please see my summary of latest changes here: https://github.com/Azure/azure-functions-powershell-worker/pull/839#issuecomment-1344568470

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,199 @@
...
+                _logger.Log(isUserOnlyLog: false, LogLevel.Trace, String.Format(
+                    PowerShellWorkerStrings.LoadingDurableSDK, PowerShellWorkerStrings.ExternalDurableSDKName));
+
+                var results = _pwsh.AddCommand(Utils.ImportModuleCmdletInfo)
+                    .AddParameter("FullyQualifiedName", PowerShellWorkerStrings.ExternalDurableSDKName)
+                    .AddParameter("ErrorAction", ActionPreference.Stop)
+                    .AddParameter("PassThru")
+                    .InvokeAndClearCommands<PSModuleInfo>();
+
+                var numResults = results.Count();
```
Francisco-Gamino: I believe that only one PowerShell module can be loaded at the time (we should test this to make sure). If this is the case, then the imported module result should always be 1. 

Also, if the result were more than one, it would be helpful if we log the module name, version, and path in Kusto for CRI investigations.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,199 @@
...
+                // If an error ocurred, we try to log the exception.
+                var errorMessage = e.ToString();
+
+                // If a PowerShell error record is available through Get-Error, we log that instead.
+                if (e.InnerException is IContainsErrorRecord inner)
+                {
+                    errorMessage = _errorRecordFormatter.Format(inner.ErrorRecord);
+
+                }
+                _logger.Log(isUserOnlyLog: true, LogLevel.Error, string.Format(
```
Francisco-Gamino: `isUserOnlyLog` should be `false` to log the entry to Kusto.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,199 @@
...
+            catch (Exception e)
+            {
+                // If an error ocurred, we try to log the exception.
+                var errorMessage = e.ToString();
+
+                // If a PowerShell error record is available through Get-Error, we log that instead.
+                if (e.InnerException is IContainsErrorRecord inner)
+                {
+                    errorMessage = _errorRecordFormatter.Format(inner.ErrorRecord);
+
```
Francisco-Gamino: Nit: please remove the extra space.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,199 @@
...
+                }
+                _logger.Log(isUserOnlyLog: true, LogLevel.Error, string.Format(
+                    PowerShellWorkerStrings.ErrorImportingDurableSDK,
+                    PowerShellWorkerStrings.ExternalDurableSDKName, errorMessage));
+
+            }
+            return importSucceeded;
+
```
Francisco-Gamino: Nit: please remove the extra space.

Francisco-Gamino (commented):
Thank you @davidmrdavid for sending this out! Looks good. I left a few comments.

src/DurableSDK/PowerShellServices.cs
davidmrdavid: Yes, I think you're right about the result always being 1. I was being overly defensive with that if-statement check. I also made a local test which validated our expectations.

I'll be removing this check then!

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,7 +352,43 @@
   <data name="DependencySnapshotDoesNotContainAcceptableModuleVersions" xml:space="preserve">
     <value>Dependency snapshot '{0}' does not contain acceptable module versions.</value>
   </data>
+  <data name="ExternalDurableSDKName" xml:space="preserve">
+    <value>AzureFunctions.PowerShell.Durable.SDK</value>
+  </data>
+  <data name="InternalDurableSDKName" xml:space="preserve">
+    <value>Microsoft.Azure.Functions.PowerShellWorker</value>
+  </data>
+  <data name="TracePipelineObjectCommand" xml:space="preserve">
```
Francisco-Gamino: This `TracePipelineObjectCommand` should not be here. You can add them as constant/variable in the class file itself, .e.g, https://github.com/Azure/azure-functions-powershell-worker/blob/ac95680f3d905a85352410d942f66e536d9ddde5/src/DependencyManagement/PowerShellGalleryModuleProvider.cs#L82 or https://github.com/Azure/azure-functions-powershell-worker/blob/ac95680f3d905a85352410d942f66e536d9ddde5/src/Utility/Utils.cs#L20

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,7 +352,43 @@
...
+    <value>Attempting to load '{0}' as the Durable Functions SDK</value>
+  </data>
+  <data name="ErrorImportingDurableSDK" xml:space="preserve">
+    <value>An unexpected error ocurred while importing '{0}': '{1}'</value>
+  </data>
   <data name="WorkerInitCompleted" xml:space="preserve">
     <value>Worker init request completed in {0} ms.</value>
   </data>
+  <data name="SetFunctionInvocationContextCmdLetTemplate" xml:space="preserve">
+    <value>{0}\Set-FunctionInvocationContext</value>
```
Francisco-Gamino: This would the be same as the comment for `TracePipelineObjectCommand`.

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,7 +352,43 @@
...
+  <data name="InternalDurableSDKName" xml:space="preserve">
+    <value>Microsoft.Azure.Functions.PowerShellWorker</value>
+  </data>
+  <data name="TracePipelineObjectCommand" xml:space="preserve">
+    <value>Microsoft.Azure.Functions.PowerShellWorker\Trace-PipelineObject</value>
+  </data>
+  <data name="UtilizingExternalDurableSDK" xml:space="preserve">
+    <value>Utilizing external Durable Functions SDK: '{0}'.</value>
+  </data>
+  <data name="isOrchestrationFailureKey" xml:space="preserve">
```
Francisco-Gamino: Nit: the first letter of the name should be capitalized.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,199 @@
...
+            var numCandidates = matchingModules.Count();
+            if (numCandidates > 0)
+            {
+                // try to import the external DF SDK
+                _usesExternalDurableSDK = tryImportingDurableSDK();
+            }
+            else
+            {
+                // Log that the module was not found in worker path
+                _logger.Log(isUserOnlyLog: false, LogLevel.Trace, string.Format(
```
Francisco-Gamino: Should we log here the values of `$env:PSModulePath`? If we decide to do it, please log the value to Kusto.

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,199 @@
...
+                var numResults = output.Count();
+                var numExpectedResults = 1;
+                var outputContractIsMet = output.Count() == numExpectedResults;
+                if (outputContractIsMet)
+                {
+                    externalInvoker = new ExternalInvoker(output[0]);
+                }
+                else
+                {
+                    var exceptionMessage = string.Format(PowerShellWorkerStrings.UnexpectedResultCount,
```
Francisco-Gamino, line 134: Should we log the module names and versions?

src/DurableSDK/PowerShellServices.cs
davidmrdavid: Yeah that seems useful, I've added it to the log message template

src/DurableSDK/PowerShellServices.cs
davidmrdavid: Yes! But we actually already do that earlier on, when we succeed in importing the DF SDK. With that in mind, do we really have to log it again?

davidmrdavid: For context, we log that here: https://github.com/Azure/azure-functions-powershell-worker/blob/7c24933ef941879b617fd69be90814cc5c3e45fe/src/DurableSDK/PowerShellServices.cs#L59-L62

src/resources/PowerShellWorkerStrings.resx
davidmrdavid: In this case, I added it inline as we only use it in one place. I think that corresponds to your second example (just like "Import-Module" is hardcoded) but perhaps I misunderstood. Please let me know if this seems right to you:

https://github.com/Azure/azure-functions-powershell-worker/blob/7c24933ef941879b617fd69be90814cc5c3e45fe/src/DurableSDK/PowerShellServices.cs#L191-L194

davidmrdavid: @Francisco-Gamino: just for my understanding, how should I determine what strings are appropiate for the PowerShellWorkerStrings.resx file?

src/resources/PowerShellWorkerStrings.resx
davidmrdavid: Sounds good, I made it in a class-level constant.

src/resources/PowerShellWorkerStrings.resx
Francisco-Gamino: Great question. The `PowerShellWorkerStrings.resx` file should only contain strings that will be displayed to the user.

Francisco-Gamino: > Great question. The `PowerShellWorkerStrings.resx` file should only contain strings that will be displayed to the user.

@davidmrdavid -- I've just realized that I was missing some information in this statement: `Great question. The PowerShellWorkerStrings.resx file should only contain strings that will be displayed to the user.` In addition to this, the strings in the resx file will be translated to all supported languages (this is because PowerShell and the help files used to be shipped with the OS), so all strings in this files will be translated. Because of this, proper names like `AzureFunctions.PowerShell.Durable.SDK` and  `Microsoft.Azure.Functions.PowerShellWorker`  should not be included these files. 
Please make these two (`Microsoft.Azure.Functions.PowerShellWorker` and `AzureFunctions.PowerShell.Durable.SDK`) constants.

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -379,11 +376,8 @@
   <data name="WorkerInitCompleted" xml:space="preserve">
     <value>Worker init request completed in {0} ms.</value>
   </data>
-  <data name="SetFunctionInvocationContextCmdLetTemplate" xml:space="preserve">
-    <value>{0}\Set-FunctionInvocationContext</value>
-  </data>
   <data name="DurableNotInWorkerPath" xml:space="preserve">
-    <value>The Durable Functions SDK, '{0}', could not be found. Please make sure this module is available in the PowerShell worker path.</value>
+    <value>The Durable Functions SDK, '{0}', could not be found. Please make sure this module is available in the PowerShell worker path. PowerShell worker path contents: '{1}'</value>
```
Francisco-Gamino: We should reference here the `PSModulePath`, e.g., this `PowerShell worker path contents: '{1}'` could be `PowerShell worker PSModulePath: '{1}'`

src/DurableSDK/PowerShellServices.cs
```diff
@@ -107,13 +99,15 @@ public void tryEnablingExternalDurableSDK()
             else
             {
                 // Log that the module was not found in worker path
+                var workerPathContents = PowerShellWorkerConfiguration.GetString("PSModulePath");
```
Francisco-Gamino: The PSModulePath is a string array, e.g., 
```powershell
PS C:\> $env:PSModulePath -split ";"
C:\Users\<userName>\OneDrive - Microsoft\Documents\PowerShell\Modules
C:\Program Files\PowerShell\Modules
c:\program files\powershell\7\Modules
C:\Program Files\WindowsPowerShell\Modules
C:\WINDOWS\system32\WindowsPowerShell\v1.0\Modules
C:\Program Files (x86)\Microsoft Azure Information Protection\Powershell
```

davidmrdavid: @Francisco-Gamino: I'm not sure I understand the feedback here. Are you saying you'd like to format the log so that each directory in PSModulePath is in its own line instead of using a single one?

Francisco-Gamino: This `PowerShellWorkerConfiguration.GetString("PSModulePath");` will return a string like this: `C:\Users\<user>\OneDrive - Microsoft\Documents\PowerShell\Modules;C:\Program Files\PowerShell\Modules;c:\program files\powershell\7\Modules;C:\Program Files\WindowsPowerShell\Modules;C:\WINDOWS\system32\WindowsPowerShell\v1.0\Modules;C:\Program Files (x86)\Microsoft Azure Information Protection\Powershell`, so you will have to format it to make it more readable.

davidmrdavid:
@Francisco-Gamino: as per our last conversation, I've refactored the code so that it no longer imports the external Durable Functions SDK. Instead, this is now the responsibility of the user app to do that: they will have to run `Import-Module` on their `profile.ps1` to import the Durable Functions SDK.

The main changes as of your latest review are in `DurableController.cs` and `PowerShellServices.cs`.

In `DurableController.cs`, all relevant logic is in `tryEnablingExternalSDK()`. This method will first check if the External DF SDK was imported onto the PowerShell session. Based on that information, and whether the customer enabled the external SDK using the environment variable`ExternalDurablePowerShellSDK`, it _may_ switch to use the external SDK and/or log warnings to the user about potential configuration errors.

This is all defined here:

https://github.com/Azure/azure-functions-powershell-worker/blob/a1e59931eeaadc9898e14938200e8b7090c68376/src/DurableWorker/DurableController.cs#L65-L92

DurableController leverages various changes in `PowerShellServices.cs`. The relevant methods are all here in the snippet below:

https://github.com/Azure/azure-functions-powershell-worker/blob/a1e59931eeaadc9898e14938200e8b7090c68376/src/DurableSDK/PowerShellServices.cs#L44-L86

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,7 +352,31 @@
...
+    <value>Unable to initialize orchestrator function due to presence of other bindings. Total number of bindings found is '{0}'. Orchestrator Functions should never use any input or output bindings other than the orchestration trigger itself. See: https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-bindings</value>
+  </data>
   <data name="WorkerInitCompleted" xml:space="preserve">
     <value>Worker init request completed in {0} ms.</value>
   </data>
+  <data name="UnexpectedResultCount" xml:space="preserve">
+    <value>Operation '{0}' expected '{1}' result(s) but received '{2}'.</value>
+  </data>
+  <data name="ExternalSDKWasNotLoaded" xml:space="preserve">
+    <value>The external Durable Functions SDK is enabled but it cannot be used because it was not loaded onto the current PowerShell session. Please add `Import-Module '{0}'` to your profile.ps1 so it may be used. Defaulting to built-in SDK.</value>
```
Francisco-Gamino: The command should be `Import-Module -Name <ModuleName> -ErrorAction Stop`

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,7 +352,31 @@
...
     <value>Dependency snapshot '{0}' does not contain acceptable module versions.</value>
   </data>
+  <data name="FoundExternalDurableSdkInSession" xml:space="preserve">
+    <value>Found external Durable Functions SDK in session: Name='{0}', Version='{1}', Path='{2}'.</value>
+  </data>
+  <data name="IsOrchestrationFailureKey" xml:space="preserve">
+    <value>IsOrchestrationFailure</value>
+  </data>
+  <data name="UnableToInitializeOrchestrator" xml:space="preserve">
+    <value>Unable to initialize orchestrator function due to presence of other bindings. Total number of bindings found is '{0}'. Orchestrator Functions should never use any input or output bindings other than the orchestration trigger itself. See: https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-bindings</value>
```
Francisco-Gamino: Would it be possible to make this link `https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-bindings` an aka link like `https://aka.ms/functions-powershell-managed-dependency`?

src/DurableSDK/PowerShellServices.cs
```diff
@@ -6,50 +6,167 @@
...
+            var numCandidates = matchingModules.Count();
+            var isModuleInCurrentSession = numCandidates > 0;
+
+            if (isModuleInCurrentSession)
+            {
+                var candidatesInfo = matchingModules.Select(module => string.Format(
+                    PowerShellWorkerStrings.FoundExternalDurableSdkInSession, module.Name, module.Version, module.Path));
+                var externalSDKModuleInfo = string.Join('\n', candidatesInfo);
+
+                if (numCandidates > 1)
```
Francisco-Gamino, line 61: Could you please add a test case that validates this logic?

src/PowerShell/PowerShellManager.cs
```diff
@@ -204,42 +205,48 @@ internal void InvokeProfile(string profilePath)
...
+                    if(functionInfo.DurableFunctionInfo.IsOrchestrationFunction)
+                    {
+                        return durableFunctionsUtils.InvokeOrchestrationFunction();
+                    }
+                    else
+                    {
+                        var isActivityFunction = functionInfo.DurableFunctionInfo.IsActivityFunction;
+                        if (!isActivityFunction)
+                        {
+                            _pwsh.AddCommand("Microsoft.Azure.Functions.PowerShellWorker\\Trace-PipelineObject");
```
Francisco-Gamino: Could you please make this `"Microsoft.Azure.Functions.PowerShellWorker\\Trace-PipelineObject"` a constant?

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,7 +352,31 @@
   <data name="DependencySnapshotDoesNotContainAcceptableModuleVersions" xml:space="preserve">
     <value>Dependency snapshot '{0}' does not contain acceptable module versions.</value>
   </data>
+  <data name="FoundExternalDurableSdkInSession" xml:space="preserve">
+    <value>Found external Durable Functions SDK in session: Name='{0}', Version='{1}', Path='{2}'.</value>
+  </data>
+  <data name="IsOrchestrationFailureKey" xml:space="preserve">
+    <value>IsOrchestrationFailure</value>
```
Francisco-Gamino: Could you please make this a constant as well? Thank you.

Francisco-Gamino (commented):
I left a couple more comments. Thank you.

src/resources/PowerShellWorkerStrings.resx
davidmrdavid: Yes, I just created `aka.ms/df-bindings`

src/DurableSDK/PowerShellServices.cs
davidmrdavid: Discussed offline, sharing update:

The methods in `PowerShellServices` aren't easy to test and given that this logic is behind a feature flag, we feel it should be safe to release as is. I have created [this](https://github.com/Azure/azure-functions-powershell-worker/issues/899) ticket to track the work of refactoring `PowerShellServices` so that unit testing these methods is easier. Prior to GA'ing this work, we will need to have unit tests for these methods.

FYI @michaelpeng36

davidmrdavid:
Thanks @Francisco-Gamino for your prompt reviews. I believe I have addressed all your comments except the unit test, to which I responded here: https://github.com/Azure/azure-functions-powershell-worker/pull/839#discussion_r1051242193

No need to merge this at this point as it won't make a difference timeline wise, although it would be good to release this with the next worker release in early January. I also won't block if people want to merge this though :) . @Francisco-Gamino please let us know if this has your approval but I'm also happy to take on more feedback as this is an important feature. Thanks all.

davidmrdavid:
Hi @Francisco-Gamino, I think in our last conversation we deemed this PR ready to merge. Please let me know if that's still the case (or let me know if you have new suggestions :) ) so that we can keep the ball rolling here and merge promptly. Thanks!

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,7 +352,28 @@
...
     <value>Worker init request completed in {0} ms.</value>
   </data>
+  <data name="UnexpectedResultCount" xml:space="preserve">
+    <value>Operation '{0}' expected '{1}' result(s) but received '{2}'.</value>
+  </data>
+  <data name="ExternalSDKWasNotLoaded" xml:space="preserve">
+    <value>The external Durable Functions SDK is enabled but it cannot be used because it was not loaded onto the current PowerShell session. Please add `Import-Module -Name {0} -ErrorAction Stop` to your profile.ps1 so it may be used. Defaulting to built-in SDK.</value>
+  </data>
+  <data name="MultipleExternalSDKsInSession" xml:space="preserve">
+    <value>Get-Module returned '{0}' instances of '{1}' in the PowerShell session, but only 1 or 0 are expected. This may create runtime errors. Please ensure your script only imports a single version of '{0}'. The modules currently in session are:\n '{2}'.</value>
```
Francisco-Gamino: Should we include the `profile.ps1` file name in this line `Please ensure your script only imports a single version of '{0}'` ?

Francisco-Gamino (approved):
I left one last minor comment. Otherwise, LGTM.

Francisco-Gamino:
@davidmrdavid -- One more thing: As per our conversation, the DF SDK will need to be imported in the `profile.ps1`. Please open an issue to track the template changes for the Functions Host, the CoreTools, and the VSCode Extension. Thank you.

src/resources/PowerShellWorkerStrings.resx
davidmrdavid: Sure, I added an explicit reference to `profile.ps1` in my latest commit: https://github.com/Azure/azure-functions-powershell-worker/pull/839/commits/8e0e103d4ffa7abd758eab6581f9e96809049533

Does that look good?

davidmrdavid:
Thank you @Francisco-Gamino.
I have created the template-tracking ticket here: ~https://github.com/Azure/azure-functions-durable-powershell-private/issues/27~ https://github.com/Azure/azure-functions-powershell-worker/issues/906
I wasn't sure what you meant by "template changes for the Functions Host" (aren't all templates technically for the Functions Host?), but I did write down that we need to modify the templates for at least core-tools and the VSCode plugin.

Please let me know if my latest edit to the log message regarding having to import the module in `profile.ps1` looks good to you. Thanks!

davidmrdavid:
FYI @Francisco-Gamino:
I'm tracking editing the `profile.ps1` templates [here](https://github.com/Azure/azure-functions-powershell-worker/issues/906), there you'll find sub-tickets for each repo where the `profile.ps1` templates are held.
And I simplified the warning log as we discussed: https://github.com/Azure/azure-functions-powershell-worker/pull/839/commits/e85c19e8070e27596d23a6af56630343d02b16ab

As the CI has passed and as per our conversation, I'll go ahead and merge this! Thank you for all your help.

PR #849, opened 2022-08-22
New programming model

andystaples:

Adds support for worker indexing to the functions worker

### Pull request checklist

### DIFF
--- a/src/RequestProcessor.cs
+++ b/src/RequestProcessor.cs
@@ -18,7 +18,10 @@ using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
 
 namespace Microsoft.Azure.Functions.PowerShellWorker
 {
+    using Microsoft.Azure.Functions.PowerShellWorker.WorkerIndexing;
+    using Microsoft.PowerShell;
     using System.Diagnostics;
+    using System.Text.Json;
     using LogLevel = Microsoft.Azure.WebJobs.Script.Grpc.Messages.RpcLog.Types.Level;
 
     internal class RequestProcessor
@@ -66,6 +69,8 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
             // If an invocation is cancelled, host will receive an invocation response with status cancelled.
             _requestHandlers.Add(StreamingMessage.ContentOneofCase.InvocationCancel, ProcessInvocationCancelRequest);
 
+            _requestHandlers.Add(StreamingMessage.ContentOneofCase.FunctionsMetadataRequest, ProcessFunctionMetadataRequest);
+
             _requestHandlers.Add(StreamingMessage.ContentOneofCase.FunctionEnvironmentReloadRequest, ProcessFunctionEnvironmentReloadRequest);
         }
 
@@ -95,6 +100,9 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
 
         internal StreamingMessage ProcessWorkerInitRequest(StreamingMessage request)
         {
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+
             var workerInitRequest = request.WorkerInitRequest;
             Environment.SetEnvironmentVariable("AZUREPS_HOST_ENVIRONMENT", $"AzureFunctions/{workerInitRequest.HostVersion}");
             Environment.SetEnvironmentVariable("POWERSHELL_DISTRIBUTION_CHANNEL", $"Azure-Functions:{workerInitRequest.HostVersion}");
@@ -115,6 +123,32 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
             {
                 RpcLogger.WriteSystemLog(LogLevel.Trace, string.Format(PowerShellWorkerStrings.SpecifiedCustomPipeName, pipeName));
                 RemoteSessionNamedPipeServer.CreateCustomNamedPipeServer(pipeName);
+            }
+
+            // Previously, this half of the dependency management would happen just prior to the dependency download in the 
+            // first function load request. Now that we have the FunctionAppDirectory in the WorkerInitRequest, 
+            // we can do the setup of these variables in the function load request. We need these variables initialized 
+            // for the FunctionMetadataRequest, should it be sent. 
+            try
+            {
+                var rpcLogger = new RpcLogger(_msgStream);
+                rpcLogger.SetContext(request.RequestId, null);
+
+                _dependencyManager = new DependencyManager(request.WorkerInitRequest.FunctionAppDirectory, logger: rpcLogger);
+
+                _powershellPool.Initialize(_firstPwshInstance);
+
+                rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, string.Format(PowerShellWorkerStrings.FirstFunctionLoadCompleted, stopwatch.ElapsedMilliseconds));
+            }
+            catch (Exception e)
+            {
+                // This is a terminating failure: we will need to return a failure response to
+                // all subsequent 'FunctionLoadRequest'. Cache the exception so we can reuse it in future calls.
+                _initTerminatingError = e;
+
+                status.Status = StatusResult.Types.Status.Failure;
+                status.Exception = e.ToRpcException();
+                return response;
             }
 
             return response;
@@ -189,26 +223,20 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
             {
                 try
                 {
-                    _isFunctionAppInitialized = true;
-
                     var rpcLogger = new RpcLogger(_msgStream);
                     rpcLogger.SetContext(request.RequestId, null);
 
-                    _dependencyManager = new DependencyManager(request.FunctionLoadRequest.Metadata.Directory, logger: rpcLogger);
+                    _isFunctionAppInitialized = true;
+
                     var managedDependenciesPath = _dependencyManager.Initialize(request, rpcLogger);
 
-                    SetupAppRootPathAndModulePath(functionLoadRequest, managedDependenciesPath);
-
-                    _powershellPool.Initialize(_firstPwshInstance);
-
+                    SetupAppRootPathAndModulePath(request.FunctionLoadRequest, managedDependenciesPath);
                     // Start the download asynchronously if needed.
                     _dependencyManager.StartDependencyInstallationIfNeeded(request, _firstPwshInstance, rpcLogger);
-
-                    rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, string.Format(PowerShellWorkerStrings.FirstFunctionLoadCompleted, stopwatch.ElapsedMilliseconds));
                 }
                 catch (Exception e)
                 {
-                    // Failure that happens during this step is terminating and we will need to return a failure response to
+                    // This is a terminating failure: we will need to return a failure response to
                     // all subsequent 'FunctionLoadRequest'. Cache the exception so we can reuse it in future calls.
                     _initTerminatingError = e;
 
@@ -341,6 +369,18 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
             return null;
         }
 
+        private StreamingMessage ProcessFunctionMetadataRequest(StreamingMessage request)
+        {
+            StreamingMessage response = NewStreamingMessageTemplate(
+                request.RequestId,
+                StreamingMessage.ContentOneofCase.FunctionMetadataResponse,
+                out StatusResult status);
+
+            response.FunctionMetadataResponse.FunctionMetadataResults.AddRange(WorkerIndexingHelper.IndexFunctions(request.FunctionsMetadataRequest.FunctionAppDirectory));
+
+            return response;
+        }
+
         internal StreamingMessage ProcessFunctionEnvironmentReloadRequest(StreamingMessage request)
         {
             var stopwatch = new Stopwatch();
@@ -393,6 +433,9 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
                     break;
                 case StreamingMessage.ContentOneofCase.FunctionEnvironmentReloadResponse:
                     response.FunctionEnvironmentReloadResponse = new FunctionEnvironmentReloadResponse() { Result = status };
+                    break;
+                case StreamingMessage.ContentOneofCase.FunctionMetadataResponse:
+                    response.FunctionMetadataResponse = new FunctionMetadataResponse() { Result = status };
                     break;
                 default:
                     throw new InvalidOperationException("Unreachable code.");
--- a/src/WorkerIndexing/BindingInformation.cs
+++ b/src/WorkerIndexing/BindingInformation.cs
@@ -1,0 +1,59 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.WorkerIndexing
+{
+    internal class BindingInformation
+    {
+        private const string BindingNameKey = "name";
+        private const string BindingDirectionKey = "direction";
+        private const string BindingTypeKey = "type";
+        public enum Directions
+        {
+            Unknown = -1,
+            In = 0,
+            Out = 1, 
+            Inout = 2
+        }
+
+        public Directions Direction { get; set; } = Directions.Unknown;
+        public string Type { get; set; } = "";
+        public string Name { get; set; } = "";
+        public Dictionary<string, Object> otherInformation { get; set; } = new Dictionary<string, Object>();
+
+        internal string ConvertToRpcRawBinding(out BindingInfo bindingInfo)
+        {
+            string rawBinding = string.Empty;
+            JObject rawBindingObject = new JObject();
+            rawBindingObject.Add(BindingNameKey, Name);
+            BindingInfo outInfo = new BindingInfo();
+
+
+            if (Direction == Directions.Unknown)
+            {
+                throw new Exception(string.Format(PowerShellWorkerStrings.InvalidBindingInfoDirection, Name));
+            }
+            outInfo.Direction = (BindingInfo.Types.Direction)Direction;
+            rawBindingObject.Add(BindingDirectionKey, Enum.GetName(typeof(BindingInfo.Types.Direction), outInfo.Direction).ToLower());
+            outInfo.Type = Type;
+            rawBindingObject.Add(BindingTypeKey, Type);
+
+            foreach (KeyValuePair<string, Object> pair in otherInformation)
+            {
+                rawBindingObject.Add(pair.Key, JToken.FromObject(pair.Value));
+            }
+
+            rawBinding = JsonConvert.SerializeObject(rawBindingObject);
+            bindingInfo = outInfo;
+            return rawBinding;
+        }
+    }
+}
--- a/src/WorkerIndexing/FunctionInformation.cs
+++ b/src/WorkerIndexing/FunctionInformation.cs
@@ -1,0 +1,40 @@
+﻿//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.WorkerIndexing
+{
+    internal class FunctionInformation
+    {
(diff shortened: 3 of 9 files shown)


andystaples:
@Francisco-Gamino I think this PR is correct. The rebase was awful and doubled the amount of commits, not sure why. Lots still to do here, gotta bundle the module with the worker and figure out how to get the cmdlet available in order to write tests for the new programming model

src/Microsoft.Azure.Functions.PowerShellWorker.csproj
```diff
@@ -35,4 +35,8 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
+  <ItemGroup>
```
Francisco-Gamino: Please remove this entry.

src/RequestProcessor.cs
```diff
@@ -153,6 +182,7 @@ internal StreamingMessage ProcessFunctionLoadRequest(StreamingMessage request)
             stopwatch.Start();
 
             FunctionLoadRequest functionLoadRequest = request.FunctionLoadRequest;
+            string lrs = System.Text.Json.JsonSerializer.Serialize<FunctionLoadRequest>(functionLoadRequest);
```
Francisco-Gamino: Could you please add a comment on what this variable is used for?

src/RequestProcessor.cs
```diff
@@ -30,9 +32,9 @@ internal class RequestProcessor
 
         // Holds the exception if an issue is encountered while processing the function app dependencies.
         private Exception _initTerminatingError;
-
-        // Indicate whether the FunctionApp has been initialized.
-        private bool _isFunctionAppInitialized;
+        
+        // Indicate whether the dependencies have been installed
+        private bool _areDependenciesInstalled;
```
Francisco-Gamino: We should keep `_isFunctionAppInitialized` given that an app could have dependencies or these might be bundled with the function app content.

In this case, `_isFunctionAppInitialized` is used to indicate whether the function app is ready for a function execution.

src/RequestProcessor.cs
```diff
@@ -117,6 +124,28 @@ internal StreamingMessage ProcessWorkerInitRequest(StreamingMessage request)
                 RemoteSessionNamedPipeServer.CreateCustomNamedPipeServer(pipeName);
             }
 
+            try
+            {
```
Francisco-Gamino, line 133: I believe that you moved the installation of function app dependencies to `ProcessWorkerInitRequest`. However, the Functions Host has a `60 seconds` timeout for the worker response, so this will cause a timeout exception. Could you please ping the Team channel on why we need this change? 

Alternately, if the function app does not have any dependencies installed, we can start an async task in `ProcessWorkerInitRequest` to install the dependencies and send the response back to the Function Host.  Then, on the first function invocation, we can check if the task completed executing, if this is the case, we can continue with function invocation. Otherwise, wait until the task completes. What are your thoughts on this @michaelpeng36 @AnatoliB @davidmrdavid ?

src/RequestProcessor.cs
```diff
@@ -240,6 +264,7 @@ internal StreamingMessage ProcessInvocationRequest(StreamingMessage request)
         {
             try
             {
+                string ins = System.Text.Json.JsonSerializer.Serialize<InvocationRequest>(request.InvocationRequest);
```
Francisco-Gamino: Could you please add a comment on what this variable is used for?

src/WorkerIndexing/BindingInformation.cs
```diff
@@ -0,0 +1,53 @@
...
+        public Directions Direction { get; set; } = Directions.Unknown;
+        public string Type { get; set; } = "";
+        public string Name { get; set; } = "";
+        public Dictionary<string, Object> otherInformation { get; set; } = new Dictionary<string, Object>();
+
+        internal string ConvertToRpcRawBinding(out BindingInfo bindingInfo)
+        {
+            string rawBinding = string.Empty;
+            JObject rawBindingObject = new JObject();
+            rawBindingObject.Add("name", Name);
```
Francisco-Gamino: Please make `"name"` a constant.

src/WorkerIndexing/FunctionInformation.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        internal RpcFunctionMetadata ConvertToRpc()
+        {
+            RpcFunctionMetadata returnMetadata = new RpcFunctionMetadata();
+            returnMetadata.FunctionId = FunctionId;
+            returnMetadata.Directory = Directory;
+            returnMetadata.EntryPoint = EntryPoint;
+            returnMetadata.Name = Name;
+            returnMetadata.ScriptFile = ScriptFile;
+            returnMetadata.Language = "powershell";
```
Francisco-Gamino: Please make `"powershell"` a constant.

src/WorkerIndexing/WorkerIndexingHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            List<RpcFunctionMetadata> indexedFunctions = new List<RpcFunctionMetadata>();
+
+            InitialSessionState initial = InitialSessionState.CreateDefault();
+            Runspace runspace = RunspaceFactory.CreateRunspace(initial);
+            runspace.Open();
+            System.Management.Automation.PowerShell ps = System.Management.Automation.PowerShell.Create();
+            ps.Runspace = runspace;
+
+            ps.AddCommand("Get-FunctionsMetadata").AddArgument(baseDir);
```
Francisco-Gamino: Please make this a constant. Here is an example: https://github.com/Azure/azure-functions-powershell-worker/blob/ac95680f3d905a85352410d942f66e536d9ddde5/src/DependencyManagement/PowerShellGalleryModuleProvider.cs#L82

src/worker.config.json
```diff
@@ -1,11 +1,12 @@
 {
-    "description":{
-        "language":"powershell",
-        "extensions":[".ps1", ".psm1"],
-        "defaultExecutablePath":"dotnet",
-        "defaultWorkerPath":"%FUNCTIONS_WORKER_RUNTIME_VERSION%/Microsoft.Azure.Functions.PowerShellWorker.dll",
-        "supportedRuntimeVersions":["7", "7.2"],
+    "description": {
```
Francisco-Gamino, line 2: We should not make any changes to the existing `worker.config.json`, so we can continue to have the existing behavior. However, could you please document the required changes in `worker.config.json` to enable `workerIndexing`? Perhaps the `src/WorkerIndexing/WorkerIndexingHelper.cs` would be a good place for this. Let me know what you think.

Francisco-Gamino (commented):
Thank you @andystaples for sending this out! I've left a few comments. Let me know if you have any questions.

src/Microsoft.Azure.Functions.PowerShellWorker.csproj
andystaples: Done

src/RequestProcessor.cs
andystaples: I was debugging with this, forgot to take it out. Removed

src/RequestProcessor.cs
andystaples: OK, changing it back

src/RequestProcessor.cs
andystaples: To clarify, the dependency download actually hasn't moved. Before and after my changes, the download happens on the first function load request. I have just moved some of the pre-initialization to the function load request, so that certain properties (like the functions app directory) are available during the metadata request which happens before function load requests. 

The worker already does kick off a task to install dependencies on the first function load request, it does not wait until all dependencies are installed to return.

src/RequestProcessor.cs
andystaples: Another debug string that I forgot to remove

src/WorkerIndexing/BindingInformation.cs
andystaples: Done

src/WorkerIndexing/FunctionInformation.cs
andystaples: Done

src/WorkerIndexing/WorkerIndexingHelper.cs
andystaples: Done

src/worker.config.json
andystaples: As per our conversation earlier this morning, two flags are necessary in order for the host to attempt worker indexing - the first is in the worker.config.json, and the second is in the local.settings.json/function app environment variables. Having this string in worker.config.json reports to the host that the worker is compatible with worker indexing, and the one in local.settings.json reports that the app is compatible. 
So, long story short, making this change doesn't harm existing apps.

src/RequestProcessor.cs
Francisco-Gamino: Thank you @andystaples for the clarification. 

Yes, in the first function invocation, we use the first PowerShell instance to download the function app dependencies and wait until the installation completes before invoking the function. The wait takes place here https://github.com/Azure/azure-functions-powershell-worker/blob/187ad3369a930f9f0acda258fa58618bf9059a27/src/DependencyManagement/DependencyManager.cs#L164

src/WorkerIndexing/WorkerIndexingHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+            ps.AddCommand(GetFunctionsMetadataCmdletName).AddArgument(baseDir);
+            string outputString = string.Empty;
+            foreach (PSObject rawMetadata in ps.Invoke())
+            {
+                if (outputString != string.Empty)
+                {
+                    throw new Exception(PowerShellWorkerStrings.GetFunctionsMetadataMultipleResultsError);
+                }
+                outputString = rawMetadata.ToString();
+                //Console.WriteLine(rawMetadata.ToString());
```
Francisco-Gamino: Please remove this if it is not being used.

src/WorkerIndexing/WorkerIndexingHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Collections.Generic;
+using System.Management.Automation;
+using System.Management.Automation.Runspaces;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.WorkerIndexing
+{
+    internal class WorkerIndexingHelper
+    {
+        const string GetFunctionsMetadataCmdletName = "Get-FunctionsMetadata";
```
Francisco-Gamino, line 19: Please include the module name, e.g., `"AzureFunctions.PowerShell.SDK\\Get-FunctionsMetadata"`

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,4 +352,10 @@
   <data name="DependencySnapshotDoesNotContainAcceptableModuleVersions" xml:space="preserve">
     <value>Dependency snapshot '{0}' does not contain acceptable module versions.</value>
   </data>
+  <data name="GetFunctionsMetadataMultipleResultsError" xml:space="preserve">
```
Francisco-Gamino, line 355: Could you please ping the Team channel to get feedback on the new string resources you added?

src/WorkerIndexing/BindingInformation.cs
```diff
@@ -0,0 +1,56 @@
...
+            outInfo.Direction = (BindingInfo.Types.Direction)Direction;
+            rawBindingObject.Add(BindingDirectionKey, Enum.GetName(typeof(BindingInfo.Types.Direction), outInfo.Direction).ToLower());
+            outInfo.Type = Type;
+            rawBindingObject.Add(BindingTypeKey, Type);
+
+            foreach (KeyValuePair<string, Object> pair in otherInformation)
+            {
+                // Wow this sucks, lots of overserialization
+                rawBindingObject.Add(pair.Key, JToken.FromObject(pair.Value));
+                //Console.WriteLine(pair.Key);
```
Francisco-Gamino: Please remove this if it is not needed.

src/WorkerIndexing/BindingInformation.cs
```diff
@@ -0,0 +1,56 @@
...
+                throw new Exception(PowerShellWorkerStrings.InvalidBindingInfoDirection);
+            }
+            outInfo.Direction = (BindingInfo.Types.Direction)Direction;
+            rawBindingObject.Add(BindingDirectionKey, Enum.GetName(typeof(BindingInfo.Types.Direction), outInfo.Direction).ToLower());
+            outInfo.Type = Type;
+            rawBindingObject.Add(BindingTypeKey, Type);
+
+            foreach (KeyValuePair<string, Object> pair in otherInformation)
+            {
+                // Wow this sucks, lots of overserialization
```
Francisco-Gamino: If you think that the overserialization issue can be improved, please add some suggestions in the comment.

src/RequestProcessor.cs
andystaples: That method waits for the dependency install task to complete, but that task is kicked off on the first function load request, using the method StartDependencyInstallationIfNeeded on line 230 of RequestProcessor.cs. Anyway, I think we're good here

src/WorkerIndexing/WorkerIndexingHelper.cs
andystaples: Done

src/WorkerIndexing/BindingInformation.cs
andystaples: Done

src/WorkerIndexing/BindingInformation.cs
andystaples: I don't have any ideas, I'm removing the comment

src/WorkerIndexing/WorkerIndexingHelper.cs
andystaples: Done

src/Microsoft.Azure.Functions.PowerShellWorker.csproj
```diff
@@ -1,4 +1,4 @@
-<!--
+﻿<!--
```
davidmrdavid: nipickt: did anything change in this file? Seems like nothing changed. If so - can we instruct git to "revert this file" so that it doesn't register a change in the PR?

src/RequestProcessor.cs
```diff
@@ -117,6 +124,28 @@ internal StreamingMessage ProcessWorkerInitRequest(StreamingMessage request)
...
+                _dependencyManager = new DependencyManager(request.WorkerInitRequest.FunctionAppDirectory, logger: rpcLogger);
+
+                _powershellPool.Initialize(_firstPwshInstance);
+
+                rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, string.Format(PowerShellWorkerStrings.FirstFunctionLoadCompleted, stopwatch.ElapsedMilliseconds));
+            }
+            catch (Exception e)
+            {
+                // Failure that happens during this step is terminating and we will need to return a failure response to
+                // all subsequent 'FunctionLoadRequest'. Cache the exception so we can reuse it in future calls.
```
davidmrdavid: A suggestion on wording:
```suggestion
                // This is a terminating failure: we will need to return a failure response to
                // all subsequent 'FunctionLoadRequest'. Cache the exception so we can reuse it in future calls.
```

src/RequestProcessor.cs
```diff
@@ -117,6 +124,28 @@ internal StreamingMessage ProcessWorkerInitRequest(StreamingMessage request)
...
+            {
+                var rpcLogger = new RpcLogger(_msgStream);
+                rpcLogger.SetContext(request.RequestId, null);
+
+                _dependencyManager = new DependencyManager(request.WorkerInitRequest.FunctionAppDirectory, logger: rpcLogger);
+
+                _powershellPool.Initialize(_firstPwshInstance);
+
+                rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, string.Format(PowerShellWorkerStrings.FirstFunctionLoadCompleted, stopwatch.ElapsedMilliseconds));
+            }
```
davidmrdavid, line 142: Could you please add a comment here on what we're trying to achieve, as well as what kind of error we're protecting against in the try-catch?

src/RequestProcessor.cs
```diff
@@ -340,6 +363,19 @@ internal StreamingMessage ProcessInvocationCancelRequest(StreamingMessage reques
...
+        
+        private StreamingMessage ProcessFunctionMetadataRequest(StreamingMessage request)
+        {
+            // WorkerStatusResponse type says that it is not used but this will create an empty one anyway to return to the host
+            StreamingMessage response = NewStreamingMessageTemplate(
+                request.RequestId,
+                StreamingMessage.ContentOneofCase.FunctionMetadataResponse,
+                out StatusResult status);
+
+            response.FunctionMetadataResponse.FunctionMetadataResults.AddRange(WorkerIndexingHelper.IndexFunctions(request.FunctionsMetadataRequest.FunctionAppDirectory));
```
davidmrdavid: Without knowing the specifics of the contract between the Host and worker, I can't decipher what the comment at the top of this block is trying to say. Can you please add a bit more context? What exactly "says that it is not used"?, and what is "it" in this context? Thanks!

src/WorkerIndexing/WorkerIndexingHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using System.Collections.Generic;
+using System.Management.Automation;
+using System.Management.Automation.Runspaces;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.WorkerIndexing
+{
+    internal class WorkerIndexingHelper
+    {
+        const string GetFunctionsMetadataCmdletName = "AzureFunctions.PowerShell.SDK\\Get-FunctionsMetadata";
```
davidmrdavid: Should we move this string to the strings-resource file?

src/WorkerIndexing/WorkerIndexingHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+            List<RpcFunctionMetadata> indexedFunctions = new List<RpcFunctionMetadata>();
+
+            InitialSessionState initial = InitialSessionState.CreateDefault();
+            Runspace runspace = RunspaceFactory.CreateRunspace(initial);
+            runspace.Open();
+            System.Management.Automation.PowerShell ps = System.Management.Automation.PowerShell.Create();
+            ps.Runspace = runspace;
+
+            ps.AddCommand(GetFunctionsMetadataCmdletName).AddArgument(baseDir);
+            string outputString = string.Empty;
```
davidmrdavid: I think this block would benefit from a small comment explaining what data your runspace is achieving, and how exactly this is helping the indexing process.

davidmrdavid (commented):
Some minor thoughts from skimming the PR. Looking good!

src/RequestProcessor.cs
andystaples: Added

src/RequestProcessor.cs
andystaples: Done

src/RequestProcessor.cs
andystaples: I have no idea what this comment is about either. My guess - some misguided note I made while working on my first prototype and forgot to remove. I have removed it

src/WorkerIndexing/WorkerIndexingHelper.cs
andystaples: This is the way that other places in the code which call cmdlets deal with it. Strings-resources are for customer-facing error/warning messages. I propose creating a separate issue on the worker to track all occurrences and move them to some global constants file.

src/WorkerIndexing/WorkerIndexingHelper.cs
andystaples: Done. This is a much larger issue, actually, and without a better approach, will lead to the AzureFunctions.PowerShell.SDK module never being available to the worker in Azure, even if we bundle it with the worker. I have left a long comment explaining the issue and some proposed solutions.

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -352,4 +352,10 @@
...
+    <value>Multiple results from metadata cmdlet</value>
+  </data>
+  <data name="InvalidBindingInfoDirection" xml:space="preserve">
+    <value>The bindingInfo's Direction is not valid</value>
```
michaelpeng36: Would we be able to specify which binding name this corresponds to?

andystaples: Yes

michaelpeng36 (approved):

src/WorkerIndexing/BindingInformation.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
```
AnatoliB: Please add the standard copyright/license header to all the new files - just copy it from other files.

andystaples: Done

PR #884, opened 2022-10-27
Populate language worker metadata in init response

### LINKED ISSUE #821: Populate WorkerMetadata in WorkerInitResponse message
The protobuf has been updated to include a `WorkerMetadata` property within the `WorkerInitResponse` message. The worker should include `WorkerMetadata` when sending the `WorkerInitResponse` to the host.

[PR](https://github.com/Azure/azure-functions-language-worker-protobuf/pull/76) for reference on how to fill in the fields.

You will need to pull in the latest protobuf version first (this [doc](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/EngineeringProcesses/ReleaseProcesses/ProtobufReleaseProcess.md&version=GBmaster&_a=preview) may help).

```proto
message WorkerMetadata {
  // The runtime/stack name
  string runtime_name = 1;

  // The version of the runtime/stack
  string runtime_version = 2;

  // The version of the worker or worker SDK
  string worker_version = 3;

  // The worker bitness/architecture
  string worker_bitness = 4;

  // Optional additional custom properties
  map<string, string> custom_properties = 5;
}
```

Francisco-Gamino:

This PR contains the following change:
- Populate the language worker metadata during worker init (resolves https://github.com/Azure/azure-functions-powershell-worker/issues/821)
 
### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/RequestProcessor.cs
+++ b/src/RequestProcessor.cs
@@ -28,6 +28,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
         private readonly System.Management.Automation.PowerShell _firstPwshInstance;
         private readonly PowerShellManagerPool _powershellPool;
         private DependencyManager _dependencyManager;
+        private string _pwshVersion;
 
         // Holds the exception if an issue is encountered while processing the function app dependencies.
         private Exception _initTerminatingError;
@@ -38,11 +39,12 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
         private Dictionary<StreamingMessage.ContentOneofCase, Func<StreamingMessage, StreamingMessage>> _requestHandlers =
             new Dictionary<StreamingMessage.ContentOneofCase, Func<StreamingMessage, StreamingMessage>>();
 
-        internal RequestProcessor(MessagingStream msgStream, System.Management.Automation.PowerShell firstPwshInstance)
+        internal RequestProcessor(MessagingStream msgStream, System.Management.Automation.PowerShell firstPwshInstance, string pwshVersion)
         {
             _msgStream = msgStream;
             _firstPwshInstance = firstPwshInstance;
             _powershellPool = new PowerShellManagerPool(() => new RpcLogger(msgStream));
+            _pwshVersion = pwshVersion;
 
             // Host sends capabilities/init data to worker
             _requestHandlers.Add(StreamingMessage.ContentOneofCase.WorkerInitRequest, ProcessWorkerInitRequest);
@@ -126,7 +128,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
                 var rpcLogger = new RpcLogger(_msgStream);
                 rpcLogger.SetContext(request.RequestId, null);
 
-                response.WorkerInitResponse.WorkerMetadata = GetWorkerMetadata(_firstPwshInstance);
+                response.WorkerInitResponse.WorkerMetadata = GetWorkerMetadata(_pwshVersion);
 
                 rpcLogger.Log(isUserOnlyLog: false, LogLevel.Trace, string.Format(PowerShellWorkerStrings.WorkerInitCompleted, stopwatch.ElapsedMilliseconds));
             }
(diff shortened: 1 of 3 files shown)


Francisco-Gamino:
@AnatoliB / @michaelpeng36 -- Could you please take a look? 
Please note that we will continue to defer the worker initialization to the first `FunctionLoadRequest` for the placeholder scenario. Also, during the first `FunctionLoadRequest` that's when we know if Managed Dependencies (MD) is enabled for the app, and if it is, we will add the MD path to the `PSModulePath`.

src/RequestProcessor.cs
```diff
@@ -117,6 +121,28 @@ internal StreamingMessage ProcessWorkerInitRequest(StreamingMessage request)
...
             }
 
+            try
+            {
+                // Initialize the first PowerShell instance, so that we can populate the worker metadata.
+                // Previously, this initialization was taking place in the first FunctionLoadRequest.
+                var rpcLogger = new RpcLogger(_msgStream);
+                rpcLogger.SetContext(request.RequestId, null);
+
+                _dependencyManager = new DependencyManager(request.WorkerInitRequest.FunctionAppDirectory, logger: rpcLogger);
```
ejizba: Just want to make sure you consider the placeholder scenario. In that case, the `FunctionAppDirectory` here will be set to a dummy app and you will later receive a `FunctionEnvironmentReloadRequest` with `FunctionAppDirectory` set to the real app

src/RequestProcessor.cs
```diff
@@ -117,6 +121,28 @@ internal StreamingMessage ProcessWorkerInitRequest(StreamingMessage request)
...
+
+                _powershellPool.Initialize(_firstPwshInstance);
```
ejizba: Another scenario to keep in mind is the multi-worker scenario. You can simulate it with this app setting:
```
"FUNCTIONS_WORKER_PROCESS_COUNT": 2
```

Alexey/Anatoli are rolling out an intelligent worker concurrency feature meaning multiple workers will happen much more often moving forward. I ran into the following issue where the second worker got a functionLoadRequest _before_ a workerInitRequest: https://github.com/Azure/azure-functions-host/issues/8825. I _think_ it only happens if worker indexing is enabled, but I'm not 100% sure.

ejizba (commented):
Hi, we perform some logic in the worker init request as well, so I have some knowledge on a few scenarios that may cause you headaches

src/RequestProcessor.cs
Francisco-Gamino: Thank you @ejizba, this is a great point. I will continue to defer the initialization until the first `FunctionLoadRequest` for the placeholder scenario.

src/RequestProcessor.cs
Francisco-Gamino: Thank you for bringing this up @ejizba. I will keep this in mind once we enable the new programming model. For the legacy programming model, everything should continue to work as expected given that the worker initialization will continue to take place during the first `FunctionLoadRequest`.

src/RequestProcessor.cs
```diff
@@ -519,6 +538,17 @@ private void SetupAppRootPathAndModulePath(FunctionLoadRequest functionLoadReque
                 .InvokeAndClearCommands();
         }
 
+        private WorkerMetadata GetWorkerMetadata(System.Management.Automation.PowerShell pwsh)
+        {
+            var data = new WorkerMetadata();
+            data.WorkerBitness = RuntimeInformation.OSArchitecture.ToString();
+            data.WorkerVersion = typeof(Worker).Assembly.GetName().Version.ToString();
+            data.RuntimeVersion = Utils.GetPowerShellVersion(pwsh);
```
AnatoliB: `GetPowerShellVersion` invocation may be quite expensive, and it requires an initialized PowerShell runspace. We are already invoking it from `Worker.Main` (via `LogPowerShellVersion`), so we'll be doing it again, and both invocations are on the cold start path. Consider storing the result of the first invocation and passing it to the `RequestProcessor` constructor.

Francisco-Gamino: Thank you Anatoli, this is really good point. I have updated the code to only get the PowerShell version once and pass it to the `RequestProcessor` constructor, so we can use it to populate the worker metadata.

src/Worker.cs
```diff
@@ -75,11 +76,16 @@ private static void WarmUpPowerShell(System.Management.Automation.PowerShell fir
...
+        private static void LogPowerShellVersion(string pwshVersion)
         {
-            var message = string.Format(PowerShellWorkerStrings.PowerShellVersion, Utils.GetPowerShellVersion(pwsh));
+            var message = string.Format(PowerShellWorkerStrings.PowerShellVersion, pwshVersion);
             RpcLogger.WriteSystemLog(LogLevel.Information, message);
         }
+
+        private static string GetPowerShellVersion(System.Management.Automation.PowerShell pwsh)
+        {
+            return Utils.GetPowerShellVersion(pwsh);
```
michaelpeng36: Could we use `Utils.GetPowershellVersion(pwsh)` instead of using this private method?

michaelpeng36 (approved):
Left one minor comment, otherwise LGTM

src/Worker.cs
Francisco-Gamino: Fixed.

AnatoliB (approved):
LGTM

PR #931, opened 2023-03-20
fix: durable orchestration with special characters

tragopogon:
Fix *Start-DurableOrchestration* with charset utf-8 to allow special characters in the *InputObject*.


davidmrdavid:
Please see my response here: https://github.com/Azure/azure-functions-powershell-worker/issues/932#issuecomment-1485998785

PR #945, opened 2023-04-21
Managed Dependencies is not supported on Legion

### LINKED ISSUE #944: Managed Dependencies is not supported on Legion
We should error out when a user tries to use Managed Dependencies on Legion.

Francisco-Gamino:

Resolves https://github.com/Azure/azure-functions-powershell-worker/issues/944

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -99,6 +99,11 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
                 // Parse and process the function app dependencies defined in the manifest.
                 _dependenciesFromManifest = _storage.GetDependencies().ToArray();
 
+                if (_dependenciesFromManifest.Length >= 1 && WorkerEnvironment.IsLinuxConsumptionOnLegion())
+                {
+                    throw new NotSupportedException(PowerShellWorkerStrings.ManagedDependenciesIsNotSupportedOnLegion);
+                }
+
                 if (!_dependenciesFromManifest.Any())
                 {
                     logger.Log(isUserOnlyLog: true, LogLevel.Warning, PowerShellWorkerStrings.FunctionAppDoesNotHaveRequiredModulesToInstall);
--- a/src/DependencyManagement/DependencyManager.cs
+++ b/src/DependencyManagement/DependencyManager.cs
@@ -99,15 +99,15 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
                 // Parse and process the function app dependencies defined in the manifest.
                 _dependenciesFromManifest = _storage.GetDependencies().ToArray();
 
-                if (_dependenciesFromManifest.Length >= 1 && WorkerEnvironment.IsLinuxConsumptionOnLegion())
-                {
-                    throw new NotSupportedException(PowerShellWorkerStrings.ManagedDependenciesIsNotSupportedOnLegion);
-                }
-
                 if (!_dependenciesFromManifest.Any())
                 {
                     logger.Log(isUserOnlyLog: true, LogLevel.Warning, PowerShellWorkerStrings.FunctionAppDoesNotHaveRequiredModulesToInstall);
                     return null;
+                }
+
+                if (WorkerEnvironment.IsLinuxConsumptionOnLegion())
+                {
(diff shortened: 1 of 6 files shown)


src/DependencyManagement/ManagedDependenciesPathDetector.cs
```diff
@@ -30,6 +31,11 @@ internal static class ManagedDependenciesPathDetector
         /// </summary>
         public static string GetManagedDependenciesPath(string functionAppRootPath)
         {
+            if (IsLinuxConsumptionOnLegion())
```
AnatoliB: Consider moving this check to a more intuitively obvious place, such as: https://github.com/Azure/azure-functions-powershell-worker/blob/25260ecd0b3f5fdb1e3a9297b7b80ec26245f80c/src/DependencyManagement/DependencyManager.cs#L107

This error should happen _only_ when the `requirements.psd1` contains any module references. The error should _not_ happen when `requirements.psd1` is absent, empty, or contains no module references. For example, a brand new app generated from a template should _not_ report this error. Chances are `GetManagedDependenciesPath` invocation does  not depend on the same factors. I suggest performing this check in `DependencyManager.Initialize`, just after we made sure that `_dependenciesFromManifest` is not empty.

src/resources/PowerShellWorkerStrings.resx
```diff
@@ -376,4 +376,7 @@
   <data name="UtilizingExternalDurableSDK" xml:space="preserve">
     <value>Utilizing external Durable Functions SDK: '{0}'.</value>
   </data>
+  <data name="ManagedDependenciesIsNotSupportedOnLegion" xml:space="preserve">
+    <value>Managed Dependencies is not supported in Linux Consumption on Legion. Please remove all module references from requirements.psd1 and include the function app dependencies with the function app content. For more information, please see https://aka.ms/functions-powershell-include-modules-with-app-content.</value>
```
AnatoliB: Consider reducing the name to just `functions-powershell-include-modules` - much shorter, but still unambiguously expressing the intent.

src/DependencyManagement/ManagedDependenciesPathDetector.cs
Francisco-Gamino: Thank you @AnatoliB for the suggestion. I have updated the code accordingly.

src/resources/PowerShellWorkerStrings.resx
Francisco-Gamino: Great suggestion. Fixed.

src/DependencyManagement/DependencyManager.cs
```diff
@@ -99,6 +99,11 @@ internal string Initialize(ILogger logger)
                 // Parse and process the function app dependencies defined in the manifest.
                 _dependenciesFromManifest = _storage.GetDependencies().ToArray();
 
+                if (_dependenciesFromManifest.Length >= 1 && WorkerEnvironment.IsLinuxConsumptionOnLegion())
```
AnatoliB: If you put this block under `if (!_dependenciesFromManifest.Any()) { ... }`, you will also be able to remove the first part of the condition.

test/Unit/DependencyManagement/DependencyManagerTests.cs
```diff
@@ -50,6 +50,38 @@ public void Initialize_ReturnsNewSnapshotPath_WhenNoAcceptableDependencyVersions
             }
         }
 
+        [Fact]
+        public void Initialize_Throws_WhenDependenciesAreDefinedInRequirementsPsd1_OnLegion()
```
AnatoliB, line 54: Consider adding a test that verifies that we _don't_ throw an exception on Legion when the dependency manifest is empty

src/DependencyManagement/DependencyManager.cs
Francisco-Gamino: Good point. Fixed.

test/Unit/DependencyManagement/DependencyManagerTests.cs
Francisco-Gamino: Test case added.

AnatoliB (approved):
LGTM

AnatoliB:
@Francisco-Gamino I think this needs to be backported to 7.2

PR #989, opened 2023-08-31
Update GH automation

bachuv:
Removing the `isAssignedToSomeone` condition when adding the `Needs: Triage` label.

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -96,7 +96,6 @@ configuration:
                         - isOpen
                         - not:
                               and:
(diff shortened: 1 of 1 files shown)


davidmrdavid (approved):

davidmrdavid:
FYI @Francisco-Gamino: there was a bug in the previous configuration where tickets would continuously get re-assigned the "needs: triage" label until someone was assigned to the ticket. With this fix, so long as we re-label the issue, we should be able to remove the "needs: triage" label

Francisco-Gamino:
Adding @Andystaples

Issue #992, opened 2023-09-08
Update language worker to support parsing command-line arguments prefix with `functions-<argumentname>`


Francisco-Gamino:
The command-line arguments that the language worker receives from the Functions Host have generic names (such as 'host' and 'port'). This can cause conflicts with other environment variables in certain environments. To address this issue, the Functions Host will now send a new set of arguments prefixed with `functions-<argumentname>`. This will be in addition to the existing arguments.

With the current parsing implementation, the language worker throws and it is not able to start given that the new prefix options are unknown, see error below:

```
ERROR(S):
  Option 'functions-uri' is unknown.
  Option 'functions-workerid' is unknown.
  Option 'functions-requestid' is unknown.
  Option 'functions-grpcmaxmessagelength' is unknown.
```
To resolve the language worker parsing issue, we need to do the following:

1) Add support for the new prefix arguments.
2) Ignore unknown arguments (this will be useful in preparation for number 3).
3) After the Functions Host changes have propagated, and we have a Core Tools version with this new Functions Host, we can remove the parsing support for the old arguments.

This is the PR that contains the Functions Host changes: https://github.com/Azure/azure-functions-host/pull/9514.

/cc @fabiocav @kshyju

fabiocav:
Could you set the parser setting to ignore unknown arguments? The command line parser supports that and it might be a better option than the current strict model. Unless you do that, you might not be able to remove the old option.

Francisco-Gamino:
This issue has been fixed.

PR #993, opened 2023-09-08
Update language worker to support parsing command-line arguments prefix with functions-<argumentname>

Francisco-Gamino:

Resolves `(1)` and `(2)` for https://github.com/Azure/azure-functions-powershell-worker/issues/992
`(3)` must be done after the Functions Host changes have propagated, and we have a Core Tools version with this new Functions Host.

Changes in this PR:
* Add support to parsing command-line arguments prefix with `functions-<argumentname>`
* Parser to ignore unknown arguments
* Remove deprecated option `grpcMaxMessageLength`

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/Worker.cs
+++ b/src/Worker.cs
@@ -31,10 +31,34 @@ namespace Microsoft.Azure.Functions.PowerShellWorker
                 LogLevel.Information,
                 string.Format(PowerShellWorkerStrings.PowerShellWorkerVersion, typeof(Worker).Assembly.GetName().Version));
 
-            WorkerArguments arguments = null;
-            Parser.Default.ParseArguments<WorkerArguments>(args)
-                .WithParsed(ops => arguments = ops)
-                .WithNotParsed(err => Environment.Exit(1));
+            var workerOptions = new WorkerOptions();
+
+            var parser = new Parser(settings => settings.EnableDashDash = true);
+            parser.ParseArguments<WorkerArguments>(args)
+                .WithParsed(workerArgs =>
+                {
+                    workerOptions.WorkerId = workerArgs.FunctionsWorkerId ?? workerArgs.WorkerId;
+                    workerOptions.RequestId = workerArgs.FunctionsRequestId ?? workerArgs.RequestId;
+
+                    if (!string.IsNullOrWhiteSpace(workerArgs.FunctionsUri))
+                    {
+                        try
+                        {
+                            var uri = new Uri(workerArgs.FunctionsUri);
+                            workerOptions.Host = uri.Host;
+                            workerOptions.Port = uri.Port;
+                        }
+                        catch (UriFormatException)
+                        {
+                            throw new ArgumentException("Invalid URI format", nameof(workerArgs.FunctionsUri));
+                        }
+                    }
+                    else
+                    {
+                        workerOptions.Host = workerArgs.Host;
+                        workerOptions.Port = workerArgs.Port;
+                    }
+                });
 
             // Create the very first Runspace so the debugger has the target to attach to.
             // This PowerShell instance is shared by the first PowerShellManager instance created in the pool,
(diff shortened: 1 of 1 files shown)


src/Worker.cs
```diff
@@ -31,10 +31,34 @@ public async static Task Main(string[] args)
...
+                    {
+                        try
+                        {
+                            var uri = new Uri(workerArgs.FunctionsUri);
+                            workerOptions.Host = uri.Host;
+                            workerOptions.Port = uri.Port;
+                        }
+                        catch (UriFormatException)
+                        {
+                            throw new ArgumentException("Invalid URI format", nameof(workerArgs.FunctionsUri));
```
kshyju: Should we also include the value of `workerArgs.FunctionsUri` in the exception so the user know what is wrong with the URL when looking at the exception message?

src/Worker.cs
```diff
@@ -31,10 +31,34 @@ public async static Task Main(string[] args)
...
+                {
+                    workerOptions.WorkerId = workerArgs.FunctionsWorkerId ?? workerArgs.WorkerId;
+                    workerOptions.RequestId = workerArgs.FunctionsRequestId ?? workerArgs.RequestId;
+
+                    if (!string.IsNullOrWhiteSpace(workerArgs.FunctionsUri))
+                    {
+                        try
+                        {
+                            var uri = new Uri(workerArgs.FunctionsUri);
+                            workerOptions.Host = uri.Host;
```
kshyju, line 56: _Nit comment unrelated to the scope of this PR change:_
You could also consider adding a Uri property to `WorkerOptions` type and assign the `workerArgs.FunctionsUri` to that. then you can avoid rebuilding the URL string inside the `MessagingStream` method. (Pass the `WorkerOptions` instances to that method)

kshyju (approved):

src/Worker.cs
```diff
@@ -31,10 +31,34 @@ public async static Task Main(string[] args)
...
-                .WithParsed(ops => arguments = ops)
-                .WithNotParsed(err => Environment.Exit(1));
+            var workerOptions = new WorkerOptions();
+
+            var parser = new Parser(settings => settings.EnableDashDash = true);
+            parser.ParseArguments<WorkerArguments>(args)
+                .WithParsed(workerArgs =>
+                {
+                    workerOptions.WorkerId = workerArgs.FunctionsWorkerId ?? workerArgs.WorkerId;
+                    workerOptions.RequestId = workerArgs.FunctionsRequestId ?? workerArgs.RequestId;
```
andystaples, line 47: This does constitute a behavior change, whereas previously the worker would exit if the WorkerId, RequestId, Host, or Port were not present, now it will continue to run (and likely crash later). Not sure if that was intended

src/Worker.cs
Francisco-Gamino: Thank you @kshyju, this is a really good suggestion. We will do this work once we remove support for parsing old command-line arguments. I have opened the following issue to track this work https://github.com/Azure/azure-functions-powershell-worker/issues/994.

src/Worker.cs
Francisco-Gamino: The intention was for the parser to be less strict (not to fail if the command line arguments contain unknown parameters). However, I do agree that we should validate the WorkerOptions before continuing execution. I have addressed that.

src/Worker.cs
Francisco-Gamino: Fixed.

andystaples (approved):

src/Worker.cs
```diff
@@ -31,10 +31,48 @@ public async static Task Main(string[] args)
                 LogLevel.Information,
                 string.Format(PowerShellWorkerStrings.PowerShellWorkerVersion, typeof(Worker).Assembly.GetName().Version));
 
-            WorkerArguments arguments = null;
-            Parser.Default.ParseArguments<WorkerArguments>(args)
-                .WithParsed(ops => arguments = ops)
-                .WithNotParsed(err => Environment.Exit(1));
+            var workerOptions = new WorkerOptions();
+
+            var parser = new Parser(settings => settings.EnableDashDash = true);
```
fabiocav: @Francisco-Gamino isn't there an property you can set using the settings provided here to allow unknown arguments? I was under the impression that was the case, and if so, setting that may simplify your change a bit.

Francisco-Gamino: Thank you @fabiocav for the suggestion. I have updated the parser settings to enable `IgnoreUnknownArguments`. I am also setting `EnableDashDash` which enables double dash '--' syntax to force parsing of all subsequent tokens as values.

src/Worker.cs
```diff
@@ -31,10 +31,52 @@ public async static Task Main(string[] args)
...
+                        {
+                            // TODO: Update WorkerOptions to have a URI property instead of host name and port number
+                            // for more information, see https://github.com/Azure/azure-functions-powershell-worker/issues/994
+                            var uri = new Uri(workerArgs.FunctionsUri);
+                            workerOptions.Host = uri.Host;
+                            workerOptions.Port = uri.Port;
+                        }
+                        catch (UriFormatException)
+                        {
+                            throw new ArgumentException($"Invalid URI format: {workerArgs.FunctionsUri}", nameof(workerArgs.FunctionsUri));
```
AnatoliB: Nit: perhaps the message from the original UriFormatException could be useful as well, consider including it into the new exception message

AnatoliB (approved):

src/Worker.cs
Francisco-Gamino: Thank you for the PR review @AnatoliB. I have updated the new exception message to include the original error message.

PR #1049, opened 2024-04-03
OpenTelemetry support in worker

### LINKED ISSUE #1048: Support OpenTelemetry in the PowerShell worker
The PowerShell worker should be able to utilize the internally owned AzureFunctions.PowerShell.OpenTelemetry.SDK module to support OpenTelemetry as configured by the user with minimal changes to the function app.

andystaples:
### Initial commit of changes needed to support OpenTelemetry in the worker
- Includes the ability to check for the presence of the AzureFunctions.PowerShell.OpenTelemetry.SDK module and run commands from that module
  - Initializes an internal activity before each function invocation to link worker traces to host
  - Takes a reference to a method from the module and forwards all logs back through this method to be passed on to the OTel endpoint
- Report the relevant OTel worker capabilities at init based on the presence of a newly specified environment variable
- Update protobuf spec to latest in order to
  - Support updating worker capabilities at EnvironmentReloadRequest
  - Provide this capability for apps which were started in placeholder mode

resolves #1048

### Pull request checklist

### Additional information

Additional PR information


andystaples:
This is a draft PR for a number of reasons. We need to discuss: 
- Any concerns with the additional checks for the presence of the OTel module
- Cold start impact
- Additional testing that may be required (?)

Also, there is a comment in there (RequestProcessor#400) that suggests some uncertainty about whether all worker capabilities need to be reported in EnvironmentReloadRequest. Reading through the protobuf spec, the host defaults to "merge" mode where new capabilities are added but old ones are retained, so that is no longer a concern. I'll pull that comment out with the next commit to this branch

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Management.Automation;
+using Microsoft.Azure.Functions.PowerShellWorker.PowerShell;
+using Microsoft.Azure.Functions.PowerShellWorker.Utility;
+using LogLevel = Microsoft.Azure.WebJobs.Script.Grpc.Messages.RpcLog.Types.Level;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.OpenTelemetrySDK
+{
+    using PowerShell = System.Management.Automation.PowerShell;
+
+    internal class OpenTelemetryController
```
AnatoliB: Quite a lot of code without unit tests.

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+    {
+        private readonly ILogger _logger;
+        private PowerShell _pwsh { get; set; }
```
AnatoliB: The direct dependency on the `PowerShell` class will complicate unit testing. Consider taking a dependency on `IPowerShellServices` instead. You may want to add either the `AddCommand` method to `IPowerShellServices`, or (probably better) more specific methods for each AddCommand/AddParameter/InvokeAndClearCommands batch.

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        public static bool isOpenTelemetryEnvironmentEnabled()
+        {
+            if (_isOpenTelmetryEnvironmentEnabled.HasValue)
+            {
+                return _isOpenTelmetryEnvironmentEnabled.Value;
+            }
+
+            string isOpenTelemetryWorkerLogEnabled = Environment.GetEnvironmentVariable(OTelEnvironmentVariableName);
+            _isOpenTelmetryEnvironmentEnabled = !string.IsNullOrEmpty(isOpenTelemetryWorkerLogEnabled);
```
AnatoliB: Is it enough to just check if the variable is present? What if it contains "FALSE", "0", etc.? @RohitRanjanMS, what is the contract?

src/Logging/RpcLogger.cs
```diff
@@ -55,6 +56,8 @@ public void Log(bool isUserOnlyLog, LogLevel logLevel, string message, Exception
                 };
 
                 _msgStream.Write(logMessage);
+
+                outputLogHandler.LogToHandlers(logLevel.ToString(), message, exception);
```
AnatoliB, line 60: Consider passing the `LogLevel` value directly, and converting it to a string later, when a string is actually needed. Hopefully, this will let you replace `Action<string, string, Exception>` with `Action<LogLevel, string, Exception>`, which is way more understandable.

src/Logging/LoggingEventHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+    internal class LoggingEventHandler
+    {
+        private HashSet<Action<string, string, Exception>> _eventHandlers = new HashSet<Action<string, string, Exception>>();
+
+        public void Subscribe(Action<string, string, Exception> handler)
+        {
+            _eventHandlers.Add(handler);
+        }
+
+        public void Unsubscribe(Action<string, string, Exception> handler)
```
AnatoliB: Doesn't seem to be used. Remove?

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+        private static bool? _isOpenTelemetryModuleLoaded;
+        private static bool? _isOpenTelmetryEnvironmentEnabled;
+
+        public OpenTelemetryController(ILogger logger, PowerShell pwsh)
+        {
+            _logger = logger;
+            _pwsh = pwsh;
+        }
+
+        public bool isOpenTelemetryCapable()
```
AnatoliB: Method naming convention: should uppercase the first character

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+    internal class OpenTelemetryController
+    {
+        private readonly ILogger _logger;
+        private PowerShell _pwsh { get; set; }
+
+        private const string StartOpenTelemetryInvocationCmdlet = "Start-OpenTelemetryInvocation";
+        private const string StopOpenTelemetryInvocationCmdlet = "Stop-OpenTelemetryInvocation";
+        private const string GetFunctionsLogHandlerCmdlet = "Get-FunctionsLogHandler";
+
+        private const string OTelEnvironmentVariableName = "OTEL_FUNCTIONS_WORKER_ENABLED";
```
AnatoliB: Consider naming it `OTelEnabledEnvironmentVariableName`.

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            if (!isOpenTelemetryCapable())
+            {
+                return;
+            }
+
+            var eventHandlers = _pwsh.AddCommand(GetFunctionsLogHandlerCmdlet)
+                .InvokeAndClearCommands<Action<string, string, Exception>>();
+
+            foreach (var eventHandler in eventHandlers)
```
AnatoliB: Do you really anticipate more than one item in `eventHandlers`? If not, perhaps you don't need a loop and a collection in `outputLogHandler`? At least not _yet_?

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+                return _isOpenTelmetryEnvironmentEnabled.Value;
+            }
+
+            string isOpenTelemetryWorkerLogEnabled = Environment.GetEnvironmentVariable(OTelEnvironmentVariableName);
+            _isOpenTelmetryEnvironmentEnabled = !string.IsNullOrEmpty(isOpenTelemetryWorkerLogEnabled);
+
+            return _isOpenTelmetryEnvironmentEnabled.Value;
+        }
+
+        public bool isOpenTelemetryModuleLoaded()
```
AnatoliB: The implementation looks very similar to `PowerShellServices.isExternalDurableSdkLoaded`. Do you think we can generalize this code instead of copying it?

src/OpenTelemetrySDK/OpenTelemetryController.cs
```diff
@@ -0,0 +1,152 @@
...
+            _isOpenTelemetryCapable = null;
+            _isOpenTelemetryModuleLoaded = null;
+            _isOpenTelmetryEnvironmentEnabled = null;
+        }
+
+        public void AddStartOpenTelemetryInvocationCommand(OpenTelemetryInvocationContext otelContext)
+        {
+            if (!isOpenTelemetryCapable() || !otelContext.isValid())
+            {
+                return;
```
AnatoliB: Does this indicate a potential problem? Perhaps we should log this condition to warn the user and us.

src/PowerShell/PowerShellManager.cs
```diff
@@ -106,6 +108,9 @@ internal void Initialize()
                 // Run the profile.ps1
                 InvokeProfile(FunctionLoader.FunctionAppProfilePath);
```
AnatoliB, line 109: It looks like we have a problem: profile invocation is not covered by OTel, correct? We don't have to solve it in this PR, but let's think about it. Eventually, we should cover it as well, as it often contains the most important diagnostics data.

src/PowerShell/PowerShellManager.cs
```diff
@@ -212,6 +218,12 @@ internal void InvokeProfile(string profilePath)
                 durableFunctionsUtils.InitializeBindings(inputData, out bool hasExternalDFsdk);
                 Logger.Log(isUserOnlyLog: false, LogLevel.Trace, String.Format(PowerShellWorkerStrings.UtilizingExternalDurableSDK, hasExternalDFsdk));
 
+                ImportEntryPointModule(functionInfo);
+
+                // This cmdlet must be run in the same .InvokeAndClearCommands() run as the function code itself
+                // If it is not, the Activity.Current context that we are trying to preserve is dropped
+                _openTelemetryController.AddStartOpenTelemetryInvocationCommand(otelContext);
```
AnatoliB, line 225: I wonder if we can invoke this before `ImportEntryPointModule`. The module may contain initialization code that runs on module import, and we should capture logs that this code emits, if possible.

src/Logging/RpcLogger.cs
andystaples: Unfortunately, this is not possible. LogLevel here comes from Microsoft.Azure.WebJobs.Script.Grpc.Messages.RpcLog.Types.Level, which is not a type that is accessible to the module. 
Even if we convert to Microsoft.Extensions.Logging.LogLevel, the worker + module will be using different versions of this assembly and therefore cannot pass objects from that assembly to each other.

src/Logging/LoggingEventHandler.cs
andystaples: I'll remove it for now - maybe worth discussing whether unsubscribing is ever necessary

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: True - this was more future-proofing than anything. I can minimize this for now

src/PowerShell/PowerShellManager.cs
andystaples: Agreed. Opened an issue from the module repo, even though the code change is likely here, just to track it together with all of the other OTel optimizations and discussions. Link: https://github.com/Azure/azure-functions-powershell-opentelemetry/issues/12

src/PowerShell/PowerShellManager.cs
andystaples: Perhaps. The old worker logic was composed of two separate calls to .InvokeAndClearCommands() - one for import the entry point module and one for the customer code. 
The call to StartOpenTelemetryInvocation must happen in the same .InvokeAndClearCommands() call as the customer code, so that the internal span is generated with the same thread ID 
We could theoretically change this logic so that all three things happen in the same thread with the same .InvokeAndClearCommands() call, but this might cause unintended changes to execution logic - for instance, if Import-Module returns some value that would be passed on in the pipeline.

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: !IsOpenTelemetryCapable() is not a problem - this is the default case for apps which are not configured to use OTel. 
!otelContext.isValid() might be worth logging - this ties into the discussion points I left in the other PR, where I am not sure what the scenario might be where the host doesn't pass TraceContext in the invocation. I'm not sure what the contract is there, I'll add some debug logging just in case

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: Currently no other workers are using this variable, so we can define the contract. I'll do bool.TryParse()

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: Refactored the code to better work with the tests, and added the very first test. More is needed before this PR can be completed

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: The structure of the Durable folders in the repo, and their associated tests, is quite a bit different (and more complex) than the current implementation of OpenTelemetry. In Durable, the DurableController is a high-level object that the other components in the worker interact with to call durable methods, but under the hood it uses it's own IPowerShellServices object (which is a durable-specific object defined in the DurableSDK folder, so I very much object to the naming there - it probably should be called IDurablePowerShellServices or maybe just IDurableServices, but what can we do) to actually do the work using the PowerShell and ILogger objects passed to it in its constructor.  

In OpenTelemetry, as it was implemented at the time of your review, the OpenTelemetryController acted as both the controller that the worker interacts with and the provider of the services underneath. 

I have refactored this - to more closely match the Controller/IServices/Services pattern, which better supports testing.

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: This is done

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: Done

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: Done

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: Minimized to single-handler set to first response from cmdlet

src/OpenTelemetrySDK/OpenTelemetryController.cs
andystaples: Done

src/OpenTelemetry/OpenTelemetryController.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
+using System.Linq;
+using Microsoft.Azure.Functions.PowerShellWorker.PowerShell;
+using Microsoft.Azure.Functions.PowerShellWorker.Utility;
+using LogLevel = Microsoft.Azure.WebJobs.Script.Grpc.Messages.RpcLog.Types.Level;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.OpenTelemetry
+{
+    using PowerShell = System.Management.Automation.PowerShell;
```
Francisco-Gamino, line 6: Are you creating a new PowerShell instance to import the module? Ideally, it would be better if we could use the first PowerShell instance to do this. This way, we wouldn't have to initialize two instances, thereby avoiding a higher cold start for this code path.

src/OpenTelemetry/OpenTelemetryInvocationContext.cs
```diff
@@ -0,0 +1,27 @@
...
+    internal class OpenTelemetryInvocationContext
+    {
+        public OpenTelemetryInvocationContext(string invocationId, string traceParent, string traceState)
+        {
+            InvocationId = invocationId;
+            TraceParent = traceParent;
+            TraceState = traceState;
+        }
+
+        public bool isValid()
```
Francisco-Gamino: Please use PascalCase convention for naming methods.

src/PowerShell/PowerShellModuleDetector.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.PowerShell
+{
+    using LogLevel = WebJobs.Script.Grpc.Messages.RpcLog.Types.Level;
+
+    internal class PowerShellModuleDetector
+    {
+        public static bool IsPowerShellModuleLoaded(System.Management.Automation.PowerShell _pwsh, ILogger _logger, string moduleName)
+        {
+            // Search for the external DF SDK in the current session
```
Francisco-Gamino: Please update this comment as the external DF SDK is one of the possible modules.

src/PowerShell/PowerShellModuleDetector.cs
```diff
@@ -0,0 +1,56 @@
...
+
+                var candidatesInfo = matchingModules.Select(module => string.Format(successMessage, module.Name, module.Version, module.Path));
+                var externalSDKModuleInfo = string.Join('\n', candidatesInfo);
+
+                if (numCandidates > 1)
+                {
+                    // If there's more than 1 result, there may be runtime conflicts
+                    // warn user of potential conflicts
+                    _logger.Log(isUserOnlyLog: false, LogLevel.Warning, String.Format(
+                        PowerShellWorkerStrings.MultipleExternalSDKsInSession,
```
Francisco-Gamino, line 44: Please update the error message `PowerShellWorkerStrings.MultipleExternalSDKsInSession` to take the name of the module we are trying to import.

andystaples: It is already there - the second string format arg takes the module name

src/OpenTelemetry/OpenTelemetryController.cs
Francisco-Gamino: This could be addressed later as a bug fix. Please open an issue and link it here.

andystaples: No, it uses the PowerShell instance that is about to execute the function. This "using" statement is because the namespace Microsoft.Azure.Functions.PowerShellWorker.PowerShell exists, and so when I try to use the type PowerShell from System.Management.Automation, it considers it ambiguous without

src/PowerShell/PowerShellModuleDetector.cs
andystaples: Done

src/OpenTelemetry/OpenTelemetryInvocationContext.cs
andystaples: Done

src/OpenTelemetry/OpenTelemetryController.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (!IsOpenTelemetryCapable())
+            {
+                return;
+            }
+
+            _services.AddStartOpenTelemetryInvocationCommand(otelContext);
+        }
+
+        public void StopOpenTelemetryInvocation(OpenTelemetryInvocationContext otelContext, bool testing = false)
```
Francisco-Gamino: Is the `testing` flag for local debugging only?

andystaples: It's for unit testing - this probably needs a longer explanation but essentially, my unit tests are checking to see if the commands were added correctly, and if you run .InvokeAndClearCommands(), there is no way to confirm that the correct commands were added

src/Logging/RpcLogger.cs
AnatoliB: Oh, I see, the handler is actually in the module, I didn't realize that. Ok, thank you.

src/PowerShell/PowerShellManager.cs
andystaples: Added a comment to the "Instrument profile.ps1" issue in the module repo that we should also ensure that ipmo for function definition file is also instrumented
https://github.com/Azure/azure-functions-powershell-opentelemetry/issues/12

src/OpenTelemetrySDK/OpenTelemetryController.cs
AnatoliB: Right, I haven't realized that IPowerShellServices is used exclusively for Durable. Creating a new interface is a good idea.

src/OpenTelemetry/OpenTelemetryServices.cs
```diff
@@ -0,0 +1,76 @@
...
+                _logger.Log(false, LogLevel.Warning, PowerShellWorkerStrings.InvalidOpenTelemetryContext);
+            }
+
+            _pwsh.AddCommand(StartOpenTelemetryInvocationCmdlet)
+                .AddParameter("InvocationId", otelContext.InvocationId)
+                .AddParameter("TraceParent", otelContext.TraceParent)
+                .AddParameter("TraceState", otelContext.TraceState);
+        }
+
+        public void StopOpenTelemetryInvocation(OpenTelemetryInvocationContext otelContext, bool _testing)
```
AnatoliB: Consider renaming `testing` to something more specific, like `addCommandOnly` or `doNotInvokeCommands`. There are so many ways to interpret the current name.

test/Unit/OpenTelemetry/OpenTelemetryControllerTests.cs
```diff
@@ -0,0 +1,163 @@
...
+        private readonly Mock<ILogger> _mockLogger = new Mock<ILogger>(MockBehavior.Strict);
+        private readonly Mock<IOpenTelemetryServices> _mockOtelServices;
+
+        public OpenTelemetryControllerTests()
+        {
+            _mockOtelServices = new Mock<IOpenTelemetryServices>(MockBehavior.Strict);
+        }
+
+        [Theory]
+        [InlineData("true", true)]
```
AnatoliB, line 29: Let's add uppercase variations here

test/Unit/OpenTelemetry/OpenTelemetryControllerTests.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            try
+            {
+                PowerShell _pwsh = PowerShell.Create();
+                var _realOTelServices = new OpenTelemetryServices(_mockLogger.Object, _pwsh);
+
+                OpenTelemetryController controller = CreateMockOpenTelemetryController();
+
+                OpenTelemetryInvocationContext context = new OpenTelemetryInvocationContext(
+                    "93d73ba2-bac9-41f9-ad31-e7ab56a6d7e1",
```
AnatoliB: These values look like magic constants. Do you have to use these specific values? Do they have to follow any specific format? If your code does not interpret them and does not really depend on them, I suggest replacing them with "fake parent id" and "fake trace id". In case the validations below ever fail, we will aprreciate the error message clarity.

src/OpenTelemetry/OpenTelemetryServices.cs
andystaples: Done

test/Unit/OpenTelemetry/OpenTelemetryControllerTests.cs
andystaples: Done

test/Unit/OpenTelemetry/OpenTelemetryControllerTests.cs
andystaples: Done

src/OpenTelemetry/IOpenTelemetryServices.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.OpenTelemetry
+{
+    internal interface IOpenTelemetryServices
```
AnatoliB: Thinking about this name, I believe the "PowerShell" part is critically important here. This interface is not about just any services required for OpenTelemetry, it is specifically about services that we expect PowerShell to provide. Consider something longer but more explicit, e.g. `IPowerShellServicesForOpenTelemetry` or something along these lines.

andystaples: Done

src/PowerShell/PowerShellManager.cs
AnatoliB: Please keep in mind that .InvokeAndClearCommands() does not create threads and does not make commands run in a separate thread. All the commands will run in the same thread that invokes .InvokeAndClearCommands(). This code:
```
pwsh.AddCommand("A");
pwsh.InvokeAndClearCommands();
pwsh.AddCommand("B");
pwsh.InvokeAndClearCommands();
```
is equivalent to this (from the _threading_ perspective):
```
pwsh.AddCommand("A");
pwsh.AddCommand("B");
pwsh.InvokeAndClearCommands();
```
Commands "A" and "B" will run with the same thread ID, and it will be the same thread that invokes AddCommand and InvokeAndClearCommands. With this in mind, I don't think we must follow this requirement: "The call to StartOpenTelemetryInvocation must happen in the same .InvokeAndClearCommands() call as the customer code".

src/PowerShell/PowerShellModuleDetector.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Linq;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Functions.PowerShellWorker.PowerShell
+{
+    using LogLevel = WebJobs.Script.Grpc.Messages.RpcLog.Types.Level;
+
+    internal class PowerShellModuleDetector
+    {
+        public static bool IsPowerShellModuleLoaded(System.Management.Automation.PowerShell _pwsh, ILogger _logger, string moduleName)
```
AnatoliB: Naming convention: we use names starting with underscore for class fields only. Parameter names should be in camelCase without underscore characters

AnatoliB: I will not block the PR on this, but please clean it up, now or later

AnatoliB (approved):
LGTM

PR #1052, opened 2024-04-11
Update azure-pipelines.yml for Azure Pipelines

Francisco-Gamino:


Francisco-Gamino:
This PR is no needed.

Issue #943, opened 2023-04-20
Update Dependency management section in the Azure Functions PowerShell developer guide


Francisco-Gamino:
Please update the **Dependency management** section at https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-powershell?tabs=portal#dependency-management. 

For dependency management, there are two options:
1) Managed dependencies (via `requirements.psd1` and `host.json`) -- these are managed by the Functions service
2) Include function app dependencies (PowerShell modules) with the function app content. These dependencies can be placed in the `Modules` folder at the function app root. This information is currently under the `Function app-level modules folder` section.

AnatoliB:
The current doc structure is outdated, mixes important concepts, and does not give enough guidance to users who do not need the Managed Dependencies feature:

- **Dependency management**
	- **Target specific versions**
		- Many users don't immediately understand that both steps (requrements.psd1 and profile.ps1) are _required_, so they discover it the hard way, when something breaks. We need to make this guidance more prominent somehow.
	- **Dependency management considerations**
	- **Dependency management app settings**
- **Custom modules**
	- **Function app-level modules folder**
		- It's surprising to see this section under _Custom modules_. This approach works with any modules, including those available on the PS Gallery
		- Not enough guidance on *how*
	- **Language worker level modules folder**
		- This section doesn't seem actionable or helpful to the user.  There are indeed modules available out-of-the-box, like with a regular PS installation, but this should be intuitively clear. The fact that this is implemented via a worker-level folder is just an implementation detail that the user cannot do anything about. I propose removing this entire section from the doc. This information is more appropriate for the design doc on the GitHub repo.
- ...
- **Considerations for PowerShell functions**
	- **Bundle modules instead of using Install-Module**
		- Bundling modules is not the only alternative to Install-Module.

I propose revamping the Dependency management section to make it very clear that there are two recommended methods to manage modules, each method has its own advantages, and each method has separate instructions. Here is the proposed structure with more comments:

- **Module management** ("Dependency management" is too ambitious - we are really talking only about modules here; it is also easy to confuse it with the Managed Dependencies as a *feature*)
	- Give a very quick comparison of two methods (using the Managed Dependencies feature vs. including modules into the app content), help the customer choose.
		- Reasons to use the Managed Dependencies feature:
			- Simplified initial installation
			- Auto-upgrades on a running app, including urgent security fixes
		- Reasons to include modules into the app content:
			- No dependency on the PowerShell Gallery
[...]

PR #1084, opened 2024-07-11
Shut off MD module upgrades after 7.4 EOL

andystaples:
Will need to be added as part of documentation work for #943 
Backport to 7.2 required, will link to that PR once this one is merged and rebased

### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR
  * [x] Otherwise: I've added my notes to `release_notes.md`
  * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr

### DIFF
--- a/src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
+++ b/src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         private bool EnableAutomaticUpgrades { get; } =
             PowerShellWorkerConfiguration.GetBoolean("MDEnableAutomaticUpgrades") ?? false;
 
+
         private readonly IDependencyManagerStorage _storage;
         private readonly IDependencySnapshotInstaller _installer;
         private readonly IDependencySnapshotPurger _purger;
@@ -45,7 +46,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
 
             _purger.SetCurrentlyUsedSnapshot(currentSnapshotPath, logger);
 
-            if (!EnableAutomaticUpgrades)
+            if (WorkerEnvironment.IsPowerShellSDKDeprecated() && !EnableAutomaticUpgrades)
             {
                 logger.Log(
                     isUserOnlyLog: false,
--- a/src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
+++ b/src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
@@ -20,6 +20,9 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.DependencyManagement
         private TimeSpan MaxBackgroundUpgradePeriod { get; } =
             PowerShellWorkerConfiguration.GetTimeSpan("MDMaxBackgroundUpgradePeriod") ?? TimeSpan.FromDays(7);
 
+        private bool EnableAutomaticUpgrades { get; } =
+            PowerShellWorkerConfiguration.GetBoolean("MDEnableAutomaticUpgrades") ?? false;
+
         private readonly IDependencyManagerStorage _storage;
         private readonly IDependencySnapshotInstaller _installer;
         private readonly IDependencySnapshotPurger _purger;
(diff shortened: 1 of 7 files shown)


test/Unit/DependencyManagement/DependencyManagerTests.cs
```diff
@@ -229,40 +229,49 @@ public void StartDependencyInstallationIfNeeded_InstallsSnapshotInForeground_Whe
         [Fact]
         public void StartDependencyInstallationIfNeeded_InvokesBackgroundMaintainer_WhenAcceptableDependenciesAlreadyInstalled()
         {
-            _mockInstalledDependenciesLocator.Setup(_ => _.GetPathWithAcceptableDependencyVersionsInstalled())
-                .Returns("AlreadyInstalled");
-            _mockStorage.Setup(_ => _.GetDependencies()).Returns(GetAnyNonEmptyDependencyManifestEntries());
+            try
+            {
+                Environment.SetEnvironmentVariable("MDEnableAutomaticUpgrades", "true");
```
AnatoliB: If we do just this, we will not have a test that verifies this behavior when the `MDEnableAutomaticUpgrades` variable is _not_ present (which is actually the most common case).

I don't have a simple solution, let's think about it. Ideally, we need to test all combinations of `MDEnableAutomaticUpgrades` (true, false, absent), both before the EOL moment and after the EOL moment, which makes it 6 test cases. But then we'll have to pass both `MDEnableAutomaticUpgrades` and a fixed "current time" to the tests. Which is doable, but perhaps there is an easier way.

src/DependencyManagement/BackgroundDependencySnapshotMaintainer.cs
```diff
@@ -42,6 +46,16 @@ public void Start(string currentSnapshotPath, DependencyManifestEntry[] dependen
 
             _purger.SetCurrentlyUsedSnapshot(currentSnapshotPath, logger);
 
+            if (WorkerEnvironment.IsPowerShellSDKDeprecated() && !EnableAutomaticUpgrades)
```
AnatoliB: This change will probably break the BackgroundDependencySnapshotMaintainer unit tests after the EOL date.

AnatoliB: Also, this implementation will make the decision just once, on a worker start. Imagine a worker running for days, weeks, or months: it will already have the `_installAndPurgeTimer` running, and nothing is going to prevent it from installing modules on the next opportunity (if I understand this code correctly). We need to find a way to make this decision more dynamic, perhaps move the check closer to the place where new modules are being installed.

AnatoliB: I think the beginning of `InstallAndPurgeSnapshots` is a better place for this check, and we can even throw an exception and let the existing catch block log it as a DependenciesUpgradeSkippedMessage warning. On the EOL date, the behavior will change without restarting the worker. I like it.

In this case, you also don't need to change `DependencyManager` or fix its unit tests anymore, as the `BackgroundDependencySnapshotMaintainer` is mocked there.

You will still need to take care of `BackgroundDependencySnapshotMaintainer` unit tests though: either provide all combinations of `MDEnableAutomaticUpgrades` and the current time, which still looks ugly. Or, much better, just add `Func<bool>` as a parameter to the `BackgroundDependencySnapshotMaintainer` constructor, so you can pass `() => WorkerEnvironment.IsPowerShellSDKDeprecated() && !EnableAutomaticUpgrades` when it's instantiated for production purposes, and `() => true` or `() => false` for test purposes. If you do this, you will have only two test cases for `BackgroundDependencySnapshotMaintainer`: one with the current behavior, and another one with the "blocked" behavior. I think we can skip testing the `WorkerEnvironment.IsPowerShellSDKDeprecated() && !EnableAutomaticUpgrades` expression itself.

andystaples: Moved to InstallAndPurgeSnapshots and made the Func<bool> change

test/Unit/DependencyManagement/DependencyManagerTests.cs
andystaples: I think there are more tests that are needed here, but wanted to get your thoughts on which ones are relevant. The test I added confirms that, if the Func<bool> passed to BackgroundDependencySnapshotMaintainer returns false, it does not attempt to install dependencies and instead warns to the logger. Any other cases you'd like me to add specific checks for?

src/DependencyManagement/WorkerEnvironment.cs
```diff
@@ -16,6 +16,8 @@ internal static class WorkerEnvironment
         private const string ContainerName = "CONTAINER_NAME";
         private const string LegionServiceHost = "LEGION_SERVICE_HOST";
 
+        private static readonly DateTime PowerShellSDKDeprecationDate = new DateTime(2026, 11, 10);
```
AnatoliB, line 19: We'll need to remember to keep updating this in the future. Perhaps we should add a reminder item to https://github.com/Azure/azure-functions-powershell-worker/blob/dev/upgrade_ps_sdk_instructions.md

test/Unit/DependencyManagement/DependencyManagerTests.cs
AnatoliB: Thank you, I think the tests we have now are adequate.

AnatoliB (approved):
LGTM

src/DependencyManagement/WorkerEnvironment.cs
andystaples: Added create_new_worker_instructions.md to document steps that must be taken when updating to a new PS 7 minor version (like 7.6). 
This file includes a note to update the Managed Dependencies shutoff date. It is likely we will find many new checklist items for this file as we start the 7.6 worker next year.

PR #1102, opened 2025-01-22
OpenAI skill trigger/Return value from pipeline

andystaples:
### Return a value for assistantSkillTrigger similar to Durable Activities

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/PowerShell/PowerShellManager.cs
+++ b/src/PowerShell/PowerShellManager.cs
@@ -243,7 +243,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.PowerShell
                     else
                     {
                         var isActivityFunction = functionInfo.DurableFunctionInfo.IsActivityFunction;
-                        if (!isActivityFunction)
+                        if (!isActivityFunction && !FunctionInfoUtilities.hasAssistantSkillTrigger(functionInfo))
                         {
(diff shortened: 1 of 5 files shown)


andystaples:
TODO: Discuss documentation, release notes, potential side-effects, parity w/ other ooproc langs

src/PowerShell/PowerShellManager.cs
```diff
@@ -247,7 +247,7 @@ public Hashtable InvokeFunction(
                         {
                             _pwsh.AddCommand(Utils.TracePipelineObjectCmdlet);
                         }
-                        return ExecuteUserCode(isActivityFunction, outputBindings);
+                        return ExecuteUserCode(isActivityFunction || FunctionInfoUtilities.hasAssistantSkillTrigger(functionInfo), outputBindings);
```
AnatoliB, line 250: Do you think we can add a test similar to `SuppressPipelineTracesForDurableActivityFunctionOnly`?

andystaples: Done

AnatoliB (approved):
Not right now, but if we need to make changes in this area in the future (perhaps adding yet another exception for other triggers): we should probably extract all the places controlling the return value and tracing behavior to a separate class, so we don't have to search for these places again and can just add one more trigger to the list. LGTM for now though.

PR #1104, opened 2025-02-27
Set PowerShell 7.4 as the default version

Francisco-Gamino:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/worker.config.json
+++ b/src/worker.config.json
@@ -5,7 +5,7 @@
         "defaultExecutablePath":"dotnet",
         "defaultWorkerPath":"%FUNCTIONS_WORKER_RUNTIME_VERSION%/Microsoft.Azure.Functions.PowerShellWorker.dll",
         "supportedRuntimeVersions":["7", "7.2", "7.4"],
-        "defaultRuntimeVersion": "7.2",
+        "defaultRuntimeVersion": "7.4",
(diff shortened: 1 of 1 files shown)


andystaples (approved):

src/worker.config.json
```diff
@@ -5,7 +5,7 @@
         "defaultExecutablePath":"dotnet",
         "defaultWorkerPath":"%FUNCTIONS_WORKER_RUNTIME_VERSION%/Microsoft.Azure.Functions.PowerShellWorker.dll",
         "supportedRuntimeVersions":["7", "7.2", "7.4"],
```
jviau, line 7: Is including 7 and 7.2 here going to be a concern as well?

Francisco-Gamino: I am not sure if we still need 7 and 7.2. @AnatoliB @andystaples - do you know?

andystaples: Good question - we are leaving these workers in the ooproc and net6 workers for now, right? 
I guess I'm not certain what effect the worker.config.json has. I know create operations from tooling use stacks API as the source of truth. I think worker.config.json has some effect with Core Tools, but again, we should be leaving the old versions in the Core Tools build, right?

Francisco-Gamino: For the time being, we will only set 7.4 as the default version.

PR #1106, opened 2025-03-14
Set environment variable to avoid Get-AzAccessToken breaking change

AnatoliB:
One of the next versions of Az.Accounts will have a breaking change: by default, the `Get-AzAccessToken` will return a `SecureString` instead a plaintext `string` even when the `-AsSecureString` switch is not provided. This environment variable will force the new `Get-AzAccessToken` to keep the old behavior by default.

### Pull request checklist

* [X] My changes **do not** require documentation changes
  * [X] Otherwise: I've added my notes to `release_notes.md`
* [X] My changes **do not** need to be backported to a previous version

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,1 +1,1 @@
-* Use rpcBaseUrl in Durable operations when possible
+* Set environment variable to avoid Get-AzAccessToken breaking change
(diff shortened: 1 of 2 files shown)


andystaples (approved):
LGTM

PR #1127, opened 2026-02-06
Enable external Durable SDK by default

andystaples:
Enables the external Durable SDK by default - throws an exception if the module is not present and the customer has not opted-out

### Pull request checklist

### DIFF
--- a/src/DurableWorker/DurableController.cs
+++ b/src/DurableWorker/DurableController.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Azure.Functions.PowerShellWorker.Durable
         private readonly ILogger _logger;
 
         private bool isExternalDFSdkEnabled { get; } =
-            PowerShellWorkerConfiguration.GetBoolean(Utils.ExternalDurableSdkEnvVariable) ?? false;
+            PowerShellWorkerConfiguration.GetBoolean(Utils.ExternalDurableSdkEnvVariable) ?? true;
 
         public DurableController(
             DurableFunctionInfo durableDurableFunctionInfo,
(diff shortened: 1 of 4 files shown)


src/DurableWorker/DurableController.cs
```diff
@@ -69,18 +69,18 @@ private void tryEnablingExternalSDK()
...
+                    // Enable external SDK (now true by default)
                     _powerShellServices.EnableExternalDurableSDK();
                 }
                 else
                 {
-                    // Customer attempted to enable external SDK but the module is not in the session. Default to built-in SDK.
-                    _logger.Log(isUserOnlyLog: false, LogLevel.Error, string.Format(PowerShellWorkerStrings.ExternalSDKWasNotLoaded, Utils.ExternalDurableSdkName));
+                    // Customer is configured to use the external SDK but the module is not in the session. Fail with helpful error message.
+                    throw new InvalidOperationException(string.Format(PowerShellWorkerStrings.ExternalSDKWasNotLoaded, Utils.ExternalDurableSdkName, Utils.ExternalDurableSdkEnvVariable));
                 }
```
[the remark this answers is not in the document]
andystaples: Can you implement the suggested check? @copilot

src/resources/PowerShellWorkerStrings.resx
AnatoliB: This message is becoming slightly confusing now. With the new setting default, in order to get it this message, the user will have to actively set `ExternalDurablePowerShellSDK` to `false`, and we should assume they had a reason to do that. In this situation, recommending them to revert it is counterproductive. Instead, we should recommend removing the external SDK. Something like this:

>Use of the built-in Durable Functions SDK is enabled by the environment variable '{1}', but '{0}' has also been imported to the PowerShell session. This may create runtime conflicts between the built-in and external Durable Functions cmdlets. If you mean to use the built-in Durable Functions SDK, please remove the '{0}' module. For more information, see https://aka.ms/durable-powershell-sdk

I've also replaced "CommandLets" with "cmdlets" (official canonic terminology).

AnatoliB: Update: added the aka.ms link

AnatoliB (changes requested):
Let's update the message I mentioned, and it will be LGTM

src/resources/PowerShellWorkerStrings.resx
andystaples: Added this change

AnatoliB (approved):

Issue #1154, opened 2026-08-05
PowerShell 7.6 external Durable SDK default still requires explicit app setting


andystaples:
## Description

PowerShell 7.6 defaults to the standalone Durable SDK in the worker, but Durable orchestrations using extension bundle 4 still fail unless the app explicitly sets `ExternalDurablePowerShellSDK=true`.

The worker default is not visible to the Durable host extension. In `Microsoft.Azure.WebJobs.Extensions.DurableTask` 3.8.0, `OrchestrationTriggerAttributeBindingProvider.OrchestrationContextToString` calls `IPlatformInformation.UsesExternalPowerShellSDK()` to choose the type-preserving history serializer. Without the explicit app setting, it uses the default serializer even though the worker selects the standalone SDK. `AzureFunctions.PowerShell.Durable.SDK` then receives history JSON without type metadata and cannot deserialize the abstract `DurableTask.Core.History.HistoryEvent` type.

## Reproduction

1. Use PowerShell worker 7.6 with extension bundle `[4.*, 5.0.0)`.
2. Add `AzureFunctions.PowerShell.Durable.SDK` 2.x to `requirements.psd1`.
3. Import it from `profile.ps1`.
4. Do not set `ExternalDurablePowerShellSDK` (rely on the PowerShell 7.6 default).
5. Start and invoke any orchestration.

The durable client starts the orchestration, but the orchestrator fails with:

```text
Could not create an instance of type DurableTask.Core.History.HistoryEvent.
Type is an interface or abstract class and cannot be instantiated.
Path 'history[0].EventType'.
```

## Expected behavior

When PowerShell 7.6 defaults to the standalone Durable SDK, the host extension should receive the same effective SDK selection and emit the type-preserving orchestration history payload without requiring a redundant app setting.

## Workaround

Explicitly set:

```text
ExternalDurablePowerShellSDK=true
```

With that setting present, the same bundle 4 / Durable SDK 2.x orchestration succeeds.

andystaples:
Resolved in AzureFunctions.PowerShell.Durable.SDK v2.4.0 - it can now parse the legacy history event types, meaning the setting is optional. Please note that this is a slightly different path than the one taken when the environment variable is present, so if experiencing issues, try adding it back as a workaround

PR #1155, opened 2026-08-05
Re-enable emulator-backed E2E tests

andystaples:
## Summary
- run the Linux E2E suite against Azurite, the Event Hubs emulator, and Cosmos DB vNext instead of live Azure resources
- update the test app for extension bundle 4, emulator-compatible client SDKs, and the standalone Durable PowerShell SDK
- add deterministic emulator readiness/cleanup and local Core Tools support
- remove the obsolete live-resource pipeline and legacy direct/nightly E2E entry point
- make `local.settings.json` explicitly emulator-backed rather than leaving Azure connection placeholders blank

## Validation
- published the PowerShell worker locally
- passed 22 emulator-backed E2E tests together (all tests except the locally shadowed managed-dependency assertion)
- passed the HTTP E2E set after the final cleanup changes
- validated PowerShell parsing, Docker Compose configuration, JSON configuration, and a clean emulator/artifact teardown

## Local test note
The managed-dependency test can fail on developer machines that already have a newer `Az.Accounts` under the user PowerShell module path. PowerShell may resolve that global copy (for example, a OneDrive Documents module installation) instead of the `ManagedDependencies` copy, even when the inherited `PSModulePath` is filtered. Clean CI agents do not have that conflicting user module, so the full suite is expected to run there. Removing or temporarily moving the user-installed module avoids the local conflict.

The standalone Durable SDK currently also requires the explicit `ExternalDurablePowerShellSDK=true` app setting with bundle 4; #1154 tracks the underlying host/worker default mismatch. This test app includes the setting as the current workaround.


Francisco-Gamino (approved):
Reviewed the emulator setup, readiness, cleanup, and CI wiring. The change looks good to me, and I like that the emulator-backed approach restores E2E coverage without requiring the previous live-resource/KeyVault path.

LGTM.

PR #1159, opened 2026-08-19
Use nuget.org vulnerability data for package audits

andystaples:
## Summary

- configure `https://data.nuget.org/v3/index.json` as the dedicated NuGet audit source
- continue restoring package content exclusively through the CFS `upstream-public` feed
- restore real vulnerability checks in CI without allowing direct package downloads from nuget.org

`data.nuget.org` exposes only vulnerability metadata, so package source isolation through CFS remains unchanged. CFS vulnerability-data support is currently in preview and can be adopted later once it is enabled for this feed.

## Validation

- ran `Check-CsprojVulnerabilities.ps1 -PrintReport` and confirmed all three project audits complete without `NU1905`
- confirmed `dotnet restore` lists only the CFS `upstream-public` feed under package feeds used
- confirmed the branch diff is limited to adding the dedicated `auditSources` configuration

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,4 +4,8 @@
     <clear />
     <add key="upstream-public" value="https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json" protocolVersion="3" />
   </packageSources>
+  <auditSources>
+    <clear />
+    <add key="upstream-public" value="https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json" protocolVersion="3" />
+  </auditSources>
 </configuration>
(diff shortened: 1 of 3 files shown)


NuGet.config
```diff
@@ -4,4 +4,8 @@
     <clear />
     <add key="upstream-public" value="https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json" protocolVersion="3" />
   </packageSources>
+  <auditSources>
+    <clear />
+    <add key="upstream-public" value="https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json" protocolVersion="3" />
```
jviau: You also are allowed to point to `data.nuget.org` for vulnerability details. That endpoint is not restricted at the moment

jviau (approved):

PR #1161, opened 2026-08-19
ci: route NuGet restores through CFS feed

jviau:
## Summary

Back-ports the CFS NuGet restore fix from `dev` commit `552dfd7` to `v4.x/ps7.6`.

- adds a root `NuGet.config` that clears inherited sources and uses only the `upstream-public` NuGet v3 feed
- adds `NuGetAuthenticate@1` to the build, test, and E2E pipeline templates

## Validation

An authenticated `dotnet restore` of the solution succeeded through the new `NuGet.config` and its sole CFS source.

Pipeline execution and the resulting `PermissiveCFSClean` telemetry cannot be verified until this change runs in Azure DevOps.

## Scope

This change deliberately addresses NuGet only. It does not change PowerShell Gallery restores because `PermissiveCFSClean` does not gate that traffic: build-correlated telemetry for definition 580 found 17 builds passing while emitting 3,284 PowerShell Gallery requests, while zero builds passed after emitting any NuGet request.

Commit `b981583` (`Route PowerShell restores through CFS`) was evaluated and deliberately not ported. The `upstream-public` feed has upstreams for npmjs, NuGet.org, PyPI, and Maven, but not PowerShell Gallery. Authenticated probes confirmed that the required PowerShell module graph is unavailable, including no result for `PSDepend`. Porting that change would therefore make builds fail deterministically during module restore rather than make them compliant. PowerShell Gallery isolation is tracked separately and requires a feed-side change first.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="upstream-public" value="https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json" protocolVersion="3" />
+  </packageSources>
+</configuration>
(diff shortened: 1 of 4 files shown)


andystaples:
Superceded by https://github.com/Azure/azure-functions-powershell-worker/pull/1160

